Issue #17, opened 2015-06-05
Consider supporting other interfaces of RestTemplate


lukasniemeier-zalando:
Right now the only interface of RestTemplate one can use with riptide is *execute* - this means one have to implement a [RequestCallback](http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/client/RequestCallback.html) when adding headers or a body, which is kind of cumbersome due to its low-level nature.

It would be great if riptide somehow leverages the magic of RestTemplate including its MessageConverters and co

Issue #16, opened 2015-05-12
Allow chained dispatch


whiskeysierra:
e.g. dispatch on status code first, and content type later

lukasniemeier-zalando:
See https://github.com/zalando/riptide/issues/7#issuecomment-109272049

PR #18, opened 2015-06-07
Feature/multi level dispatch

### LINKED ISSUE #7: Add support for "default dispatch"

whiskeysierra:
Would fix #7, #15, #16 and #17


whiskeysierra:
No idea why coveralls thinks [this line](https://coveralls.io/files/825702587#L72) is uncovered. It's the only call to a private method that in fact has coverage... :/

PR #21, opened 2015-06-15
Added support for layered propagation

### LINKED ISSUE #20: Propagate response to upper level if no match found
Assuming you get a 200 appication/json back and this is your routing table:

    2xx
      200
        application/xml
        text/plain
      201
      any: onAnySuccess
    4xx
    5xx

Since there is no specific match for that content type in `2xx > 200` riptide should fall back to the `onAnySuccess` handler (or even higher up the chain) if needed.

whiskeysierra:
Fixes #20


lukasniemeier-zalando:
+1

PR #22, opened 2015-06-18
OAuth2 Compatibility: Propagation of Response

lukasniemeier-zalando:
Added a special _OAuth2CompatibilityResponseErrorHandler_ which ensure correct dispatching when using a _OAuth2RestTemplate_.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,11 +23,21 @@ differently with an easy to use yet very powerful syntax.
 
 ## Usage
 
-Create an instance based on an existing `RestTemplate` (also see section *RestTemplate Configuration*):
+Create an instance based on an existing `RestTemplate`:
 
 ```java
-final Rest rest = Rest.create(new RestTemplate());
+final RestTemplate template = new RestTemplate();
+template.setResponseErrorHandler(new PassThroughResponseErrorHandler());
+final Rest rest = Rest.create(template);
 ```
+
+If you use Riptide to its full extent you probably don't want to have any [`ResponseErrorHandler`]
+(http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/client/ResponseErrorHandler.html)
+interfere with your dispatching. Therefore Riptide provides you with a *null* `ResponseErrorHandler`, which ensures
+that Riptide handles all success and error cases.
+
+**When using OAuth2RestTemplate** you have to use the `OAuth2CompatibilityResponseErrorHandler`, which ensures that
+dispatching works even if OAuth errors occur.
 
 Make a request and route the response to your specific handler methods/callbacks:
 
@@ -154,21 +164,6 @@ final Success success = rest.execute(GET, url)
                 anySeries().call(this::fail))
         .retrieve(Success.class).orElse(null);
 ```
-
-### RestTemplate Configuration
-
-If you use Riptide to its full extent you probably don't want to have any [`ResponseErrorHandler`]
-(http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/client/ResponseErrorHandler.html)
-interfere with your dispatching. Therefore Riptide provides you with a *null* `ResponseErrorHandler`.
-
-```java
-final RestTemplate template = new RestTemplate();
-template.setResponseErrorHandler(new PassThroughResponseErrorHandler());
-final Rest rest = Rest.create(template);
-```
-
-**When using OAuth2RestTemplate** you have to use the `OAuth2CompatibilityResponseErrorHandler`, which ensures that
-dispatching works even if OAuth errors occur.
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ differently with an easy to use yet very powerful syntax.
 
 ## Usage
 
-Create an instance based on an existing `RestTemplate`:
+Create an instance based on an existing `RestTemplate` (also see section *RestTemplate Configuration*):
 
 ```java
 final Rest rest = Rest.create(new RestTemplate());
@@ -154,6 +154,21 @@ final Success success = rest.execute(GET, url)
                 anySeries().call(this::fail))
         .retrieve(Success.class).orElse(null);
 ```
+
+### RestTemplate Configuration
+
+If you use Riptide to its full extent you probably don't want to have any [`ResponseErrorHandler`]
+(http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/client/ResponseErrorHandler.html)
+interfere with your dispatching. Therefore Riptide provides you with a *null* `ResponseErrorHandler`.
+
+```java
+final RestTemplate template = new RestTemplate();
+template.setResponseErrorHandler(new PassThroughResponseErrorHandler());
+final Rest rest = Rest.create(template);
+```
+
+**When using OAuth2RestTemplate** you have to use the `OAuth2CompatibilityResponseErrorHandler`, which ensures that
+dispatching works even if OAuth errors occur.
 
 ## License
 
--- a/pom.xml
+++ b/pom.xml
@@ -97,6 +97,13 @@
             <version>${jackson.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.security.oauth</groupId>
+            <artifactId>spring-security-oauth2</artifactId>
+            <version>2.0.7.RELEASE</version>
+            <scope>test</scope>
+            <optional>true</optional>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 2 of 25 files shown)


README.md
```diff
@@ -23,7 +23,7 @@ differently with an easy to use yet very powerful syntax.
 
 ## Usage
 
-Create an instance based on an existing `RestTemplate`:
+Create an instance based on an existing `RestTemplate` (also see section *RestTemplate Configuration*):
```
whiskeysierra: You can use a real link:

    [*RestTemplate Configuration*](#resttemplate-configuration)

src/main/java/org/zalando/riptide/handler/OAuth2CompatibilityResponseErrorHandler.java
```diff
@@ -0,0 +1,41 @@
...
+ * ​⁣
+ */
+
+import org.springframework.http.client.ClientHttpResponse;
+import org.springframework.web.client.ResponseErrorHandler;
+import org.zalando.riptide.AlreadyConsumedResponseException;
+
+import java.io.IOException;
+
+public final class OAuth2CompatibilityResponseErrorHandler implements ResponseErrorHandler {
```
whiskeysierra: A class comment would help to understand the need for this class. It could incorporate the relevant code pieces from the OAuthResponseErrorHandler to clarify the issue it solves.

src/main/java/org/zalando/riptide/handler/OAuth2CompatibilityResponseErrorHandler.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public boolean hasError(ClientHttpResponse response) throws IOException {
+        return false;
+    }
+
+    @Override
+    public void handleError(ClientHttpResponse response) throws IOException {
+        throw new AlreadyConsumedResponseException(response);
```
whiskeysierra: To clarify: this works because a) we decided it's not an error (see line 33) but *someone*(tm) decided to call us anyways; we throw an exception in which we capture the given response and propagate it up the call stack until we catch it again (in Rest.java#64) were we treat it like a normal execution path (which it really should be) and start normal dispatching on it. Fundamental requirement is that the OAuth code is not doing anything after calling the underlying response handler.

src/test/java/org/zalando/riptide/OAuth2CompatibilityTest.java
```diff
@@ -0,0 +1,95 @@
...
+    @Test
+    public void dispatchesConsumedResponseAgain() throws IOException {
+        server.expect(requestTo(url)).andRespond(withSuccess());
+
+        Rest rest = Rest.create(template);
+        rest.execute(GET, url).dispatch(status(), on(HttpStatus.OK).capture()).retrieveResponse()
+            .orElseThrow(() -> new AssertionError("Expected a response"));
+    }
+
+    private static class OAuth2ResponseErrorHandler implements ResponseErrorHandler {
```
whiskeysierra: Couldn't we try to use the real thing just for the tests (`<scope>test</scope>`) or would it be too much hassle to make it work?

src/test/java/org/zalando/riptide/OAuth2CompatibilityTest.java
```diff
@@ -0,0 +1,95 @@
...
+        template.setErrorHandler(new OAuth2ResponseErrorHandler(handler));
+
+        server = MockRestServiceServer.createServer(template);
+    }
+
+    @Rule
+    public ExpectedException exception = ExpectedException.none();
+
+    @Test
+    public void dispatchesConsumedResponseAgain() throws IOException {
```
whiskeysierra: I feel like there should be a bad case test as well; one we don't use our compatibility handler and it fails.

src/test/java/org/zalando/riptide/PassThroughResponseErrorHandlerTest.java
```diff
@@ -0,0 +1,42 @@
...
+    @Test
+    public void isNoErrorForServerError() throws IOException {
+
+
+        assertThat(unit.hasError(new MockClientHttpResponse(new byte[]{}, HttpStatus.INTERNAL_SERVER_ERROR)),
+                is(false));
+    }
+
+    @Test
+    public void throwsResponseWrappedInException() throws IOException {
```
whiskeysierra: The name is a bit misleading; it's not throwing anything, is it?

src/test/java/org/zalando/riptide/OAuth2CompatibilityTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+        template = new RestTemplate();
+        template.setMessageConverters(singletonList(converter));
+        template.setErrorHandler(new OAuth2ResponseErrorHandler(handler));
+
+        server = MockRestServiceServer.createServer(template);
+    }
+
+    @Rule
+    public ExpectedException exception = ExpectedException.none();
```
whiskeysierra: `final`

README.md
whiskeysierra: Thinking about it, I'd like to see the setup code with the `PassThroughResponseErrorHandler` up here. The OAuth2 stuff could be its own section below (and linked from here).

whiskeysierra: > Thinking about it, I'd like to see the setup code with the PassThroughResponseErrorHandler up here.

Especially important as the first usage example as it stands doesn't work out of the box (due to the default response error handler).

src/main/java/org/zalando/riptide/AlreadyConsumedResponseException.java
```diff
@@ -0,0 +1,36 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ​⁣
+ */
+
+import org.springframework.http.client.ClientHttpResponse;
+
+public class AlreadyConsumedResponseException extends RuntimeException {
```
whiskeysierra: `final`

src/main/java/org/zalando/riptide/AlreadyConsumedResponseException.java
```diff
@@ -0,0 +1,36 @@
...
+
+    private final ClientHttpResponse response;
+
+    public AlreadyConsumedResponseException(final ClientHttpResponse response) {
```
whiskeysierra: All package private?

README.md
lukasniemeier: +1 I'll move the explanation and initialization example up

src/main/java/org/zalando/riptide/handler/OAuth2CompatibilityResponseErrorHandler.java
lukasniemeier: Will use your summary as a base for the class comment :soon:

src/test/java/org/zalando/riptide/OAuth2CompatibilityTest.java
lukasniemeier: I'll try what I need for such an integration test. Will mark it as ``<optional>true</optional>`` also.

src/test/java/org/zalando/riptide/PassThroughResponseErrorHandlerTest.java
lukasniemeier: Name is outdated, thank's!

src/main/java/org/zalando/riptide/AlreadyConsumedResponseException.java
lukasniemeier: :+1:

src/test/java/org/zalando/riptide/OAuth2CompatibilityTest.java
whiskeysierra: > Will mark it as <optional>true</optional> also.

Is that needed for tests? Would it make a difference?

lukasniemeier: IIRC this will prevent this dependency to be resolved as a transitive dependency when someone is using Riptide (in the test-scope).

pom.xml
```diff
@@ -97,6 +97,13 @@
             <version>${jackson.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.security.oauth</groupId>
+            <artifactId>spring-security-oauth2</artifactId>
+            <version>2.0.7.RELEASE</version>
+            <scope>test</scope>
+            <optional>true</optional>
```
whiskeysierra: According to [MNG-1378](https://issues.apache.org/jira/browse/MNG-1378):
> The problem here is that by design test dependencies are not transitive

By default they are not even installed/deployed so no one can depend on them, afaik.

PR #28, opened 2015-08-10
Optional on typedcondition call

### LINKED ISSUE #25: TypedCondition::call should use Optional
Right now ``TypedCondition::call`` expects a ``EntityConsumer<I>`` when creating a ``Binding`` for type ``I``. I think it would be better if the consumer is of ``EntityConsumer<Optional<I>>``.

Looking at ``HttpMessageConverterExtractor`` there are some cases where the ``entity`` will be ``null`` (e.g. empty response body for a *HEAD* request). Client's should think about such cases when using ``TypedCondition::call``.

lukasniemeier-zalando:
May resolve #25 

MessageConverters use *null* as a valid value on deserializing a type.
As we already wrap those values in an Optional for the *Receiver*
interface this commit adds this to the TypedCondition API.

This change breaks current API usages.

### DIFF
--- a/src/main/java/org/zalando/riptide/TypedCondition.java
+++ b/src/main/java/org/zalando/riptide/TypedCondition.java
@@ -57,7 +57,7 @@ public final class TypedCondition<A, I> implements Capturer<A> {
     public Binding<A> call(EntityConsumer<Optional<I>> consumer) {
         return Binding.create(attribute, (response, converters) -> {
             final I entity = convert(response, converters);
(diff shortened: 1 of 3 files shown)


whiskeysierra:
If this only affects `HEAD` requests I'd argue that we add better support for capturing and retrieving headers. Thoughts?

whiskeysierra:
I added better support to retrieve headers for `HEAD` requests.

PR #36, opened 2015-08-19
Feature/exception handling

### LINKED ISSUE #35: Reconsider exception handling concept
Right now the exception handling is kind of shady:

- `any*().call(..)` handles the *no match case* (which is fine)
- it also catches and swallows any `RestClientException` and `HttpMessageNotReadableException`
- the general idea should be: pass through everything *as-is* from custom callbacks/mappers as well as from the underlying RestTemplate
- the only case I can think of right now that justifies a custom exception is *no match* without an *any* condition

whiskeysierra:
Fixes #35


whiskeysierra:
Merge after #34

PR #51, opened 2016-03-29
Added support for AsyncRestTemplate

whiskeysierra:
It's still missing one crucial part: We need to define a failure callback for errors during asynchronous response handling. Two possibilities:

1. define failure callback globally
2. define failure callback as an additional parameter to all `AsyncRest.execute` methods

@lukasniemeier-zalando What do you think?


whiskeysierra:
It's also missing some changes to the readme, but I'd wait for #50 first.

lukasniemeier-zalando:
I'd vote for solution (2) - similar to the `then` `catch` `finally` pattern.

lukasniemeier:
:+1: on merge (but adding a section to the README explaining the new syntactic sugar for async might be good).

PR #60, opened 2016-05-14
Added support for normalizing responses (Location/Content-Location header)

### LINKED ISSUE #52: Should we resolve relative Location/Content-Location headers automatically?

whiskeysierra:
Fixes #52

### DIFF
--- a/src/main/java/org/zalando/riptide/Actions.java
+++ b/src/main/java/org/zalando/riptide/Actions.java
@@ -24,8 +24,13 @@ import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMessage;
 import org.springframework.http.client.ClientHttpResponse;
 
+import javax.annotation.Nullable;
 import java.io.IOException;
 import java.net.URI;
+import java.util.Optional;
+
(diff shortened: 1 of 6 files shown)


src/main/java/org/zalando/riptide/Actions.java
```diff
@@ -48,10 +53,53 @@
...
+     *
+     * @param uri the base uri to resolve against
+     * @return a function that normalizes responses
+     */
+    public static ThrowingFunction<ClientHttpResponse, ClientHttpResponse, IOException> normalize(final URI uri) {
+        return response -> {
+            final HttpHeaders headers = new HttpHeaders();
+            headers.putAll(response.getHeaders());
+
+            Optional.ofNullable(headers.getLocation())
```
jhorstmann: Would it make sense to first check whether we received a relative location or content-location? Maybe it's not needed, when the client uses this function it probably already knows that the server will use a relative uri.

whiskeysierra: > when the client uses this function it probably already knows that the server will use a relative uri.

In my opinion the client can only know for sure by looking at the response.

PR #61, opened 2016-05-14
Added support for Future<Capture>

### LINKED ISSUE #59: Allow AsyncRest's Capture to return Futures

whiskeysierra:
Fixes #59 

The only downside to this is that changing a void method to a return type is breaking change (see http://stackoverflow.com/questions/3589946/retrofitting-void-methods-to-return-its-argument-to-facilitate-fluency-breaking/3589948#3589948). So this would require to go for Riptide 2.0.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,8 +167,7 @@ response will be routed to an action. An action can be one of the following type
 Consumers can be used to trigger some dedicated function and they work well if no return value is required.
 
 Functions on the other hand are used to apply a transformation and their result must be captured. Captured values can 
-later be retrieved, e.g. to produce a return value. Please be aware that captures are not available when using
-`AsyncRest`.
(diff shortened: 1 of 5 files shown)


tkrop:
:+1:

jhorstmann:
> The only downside to this is that changing a void method to a return type is breaking change

Would it be worth it to add this as a separate method like `dispatchAsync` to keep compatibility?

On the bytecode level it would even be possible to add a method with same name and different return type, but `javac` won't allow you to do that :)

whiskeysierra:
> Would it be worth it to add this as a separate method like dispatchAsync to keep compatibility?

Feels like smurf-naming, since the class is already called `AsyncRest`

whiskeysierra:
Closing this because we put this onto a release branch

Issue #56, opened 2016-05-02
Allow call() with a parameterless "consumer" (e.g.. Runnable)


ePaul:
There are some cases where the conditions of the dispatch function are already everything which is interesting (i.e. neither response body nor any remaining header information matter). (An example is the `200` response from Nakadi's "submit events" operation.)

But currently I still need to pass to `.call()` a consumer of either some class mapped from the response, or even the HttpClientResponse.

I propose to add a `call` method in UntypedCondition which accepts a `Runnable` (or a throwing version thereof), simply discarding the ClientHttpResponse and calling the runnable.

For now I'm using this `wrap` function as a workaround:

```java
    private static ThrowingConsumer<ClientHttpResponse, ?> wrap(Runnable runnable) {
        return response -> runnable.run();
    }
```

whiskeysierra:
This addition would be especially interesting in regards of support for AsyncRestTemplate, where any code that needs to run after a response needs to be supplied as some kind of callback. (Compared to be just placed after the call to `rest.execute(..)`.

tkrop:
@whiskeysierra @ePaul is there semantic difference of this suggestion to `rest.execute(GET, url).dispatch(status(),on(UNAUTHORIZED).capture())`?

ePaul:
@tkrop I'm not sure I understand your question.

The point of `call(...)` is to specify code to execute, with `capture()` you say "store something for later retrieval". (Though I'm not sure what is actually captured in your example.)

Currently I can say 

```java
rest.execute(GET, url).dispatch(status(),
          on(OK).call(response -> callback.successfullyPublished()));
```
But this response is not actually used in the callback (it's a void method without parameters), thus I would like to be able to say
```java
rest.execute(GET, url).dispatch(status(),
          on(OK).call(callback::successfullyPublished));
```
directly. This is the difference between a `Consumer<ClientHttpResponse>` and a `Runnable`.

whiskeysierra:
@ePaul Are you willing to do a pull-request?

ePaul:
I'll have a try, can't promise it will work.

whiskeysierra:
I can't think of a reason why it shouldn't.

ePaul:
First thing going wrong is that my home computer has no Java 8, and the Ubuntu is old enough there isn't even a package to install it. So, it might take a while until I set up everything. (Also, 4 GB seems to not be enough to run Eclipse and Opera with several tabs open at the same time.)

PR #63, opened 2016-05-17
Feature/#56 allow call with parameterless function

ePaul:
Attempt to implement #56.

This adds a `.call(...)` overload to both TypedCondition and UntypedCondition, accepting a ThrowingRunnable each.

I'm not sure the addition to TypedCondition is actually sensible (because why map to a type if you are then throwing it away?), please comment on this.

### DIFF
--- a/src/test/java/org/zalando/riptide/AsyncTest.java
+++ b/src/test/java/org/zalando/riptide/AsyncTest.java
@@ -75,6 +75,20 @@ public final class AsyncTest {
     }
 
     @Test
+    public void shouldCallWithoutParameters() {
+        server.expect(requestTo(url)).andRespond(withSuccess());
+
+        @SuppressWarnings("unchecked")
+        final ThrowingRunnable<RuntimeException> verifier = mock(ThrowingRunnable.class);
+
+        unit.execute(GET, url).dispatch(series(),
+                on(SUCCESSFUL).call(verifier));
+
+        verify(verifier).run();
+    }
+
+
+    @Test
     public void shouldCallWithHeaders() {
         server.expect(requestTo(url)).andRespond(withSuccess());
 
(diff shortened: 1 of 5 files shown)


whiskeysierra:
> I'm not sure the addition to TypedCondition is actually sensible (because why map to a type if you are then throwing it away?), please comment on this.

I agree. I'd only add this to UntypedCondition.

src/test/java/org/zalando/riptide/AsyncTest.java
```diff
@@ -75,6 +75,20 @@ public void shouldCall() {
     }
 
     @Test
+    public void shouldCallWithoutParameters() {
```
whiskeysierra: Why is this in the AsyncTest?

ePaul: Because this was the only point I've found the other untyped `call` method being tested.

ePaul:
> I agree. I'd only add this to UntypedCondition.

Removed from TypedCondition.

src/test/java/org/zalando/riptide/AsyncTest.java
whiskeysierra: Now we have the same test twice.

ePaul: Though I guess it doesn't really fit at this place ... I've misread the other `shouldCall...` test methods there.

I still don't find any place the previous UntypedCondition.call is explicitly tested (it is used in other tests to see which of the conditions matched).

ePaul: > Now we have the same test twice.

Once with a synchronous Rest, once with AsyncRest, yes.
I guess I can remove it from here if you don't deem it necessary to test that this works asynchronously, too.

whiskeysierra: You're right, it's testing two different things, actually.

whiskeysierra:
:+1:

PR #68, opened 2016-05-17
Feature/urltemplate

AlexanderYastrebov:
- [x] document use case

### DIFF
--- a/src/main/java/org/zalando/riptide/RestBase.java
+++ b/src/main/java/org/zalando/riptide/RestBase.java
@@ -1,0 +1,68 @@
+package org.zalando.riptide;
+
+/*
+ * ⁣​
+ * Riptide
+ * ⁣⁣
+ * Copyright (C) 2015 Zalando SE
+ * ⁣⁣
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ​⁣
+ */
+
+import java.net.URI;
+import java.util.Map;
+import java.util.function.Supplier;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.web.util.UriTemplateHandler;
+
+abstract class RestBase<TEMPLATE, DISPATCHER> {
+
+    protected final Router router = new Router();
+    protected final TEMPLATE template;
+
+    private final Supplier<UriTemplateHandler> uriTemplateHandler;
+
+    protected RestBase(final TEMPLATE template, final Supplier<UriTemplateHandler> uriTemplateHandler) {
+        this.template = template;
+        this.uriTemplateHandler = uriTemplateHandler;
+    }
+
+    protected abstract <T> DISPATCHER execute(HttpMethod method, URI url, HttpEntity<T> entity);
+
+    public RestWithURL<DISPATCHER> withUrl(final String uriTemplate, final Object... uriVariables) {
+        return new RestWithURL<>(this, uriTemplateHandler.get().expand(uriTemplate, uriVariables));
+    }
+
+    public RestWithURL<DISPATCHER> withUrl(final String uriTemplate, final Map<String, ?> uriVariables) {
+        return new RestWithURL<>(this, uriTemplateHandler.get().expand(uriTemplate, uriVariables));
+    }
+
+    public DISPATCHER execute(final HttpMethod method, final URI url) {
+        return execute(method, url, HttpEntity.EMPTY);
+    }
+
+    public DISPATCHER execute(final HttpMethod method, final URI url, final HttpHeaders headers) {
+        return execute(method, url, new HttpEntity<>(headers));
+    }
+
+    public DISPATCHER execute(final HttpMethod method, final URI url, final Object body) {
+        return execute(method, url, new HttpEntity<>(body));
+    }
+
+    public DISPATCHER execute(final HttpMethod method, final URI url, final HttpHeaders headers, final Object body) {
+        return execute(method, url, new HttpEntity<>(body, headers));
+    }
+}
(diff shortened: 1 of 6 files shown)


src/main/java/org/zalando/riptide/RestBase.java
```diff
@@ -0,0 +1,68 @@
...
+
+import java.net.URI;
+import java.util.Map;
+import java.util.function.Supplier;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.web.util.UriTemplateHandler;
+
+abstract class RestBase<TEMPLATE, DISPATCHER> {
```
whiskeysierra: `<T, D>`?

src/main/java/org/zalando/riptide/RestWithURL.java
```diff
@@ -0,0 +1,53 @@
...
+ * limitations under the License.
+ * ​⁣
+ */
+
+import java.net.URI;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+
+public class RestWithURL<DISPATCHER> {
```
whiskeysierra: I like the concept of the class, but we have to find a better name.

src/main/java/org/zalando/riptide/RestBase.java
AlexanderYastrebov: But this is more readable despite unusual

src/main/java/org/zalando/riptide/RestWithURL.java
AlexanderYastrebov: Also this name supposed to be invisible to client as he will immediately chain execute call

src/main/java/org/zalando/riptide/RestBase.java
```diff
@@ -0,0 +1,68 @@
...
+
+    protected final Router router = new Router();
+    protected final TEMPLATE template;
+
+    private final Supplier<UriTemplateHandler> uriTemplateHandler;
+
+    protected RestBase(final TEMPLATE template, final Supplier<UriTemplateHandler> uriTemplateHandler) {
```
whiskeysierra: Couldn't we just extract this from the template, instead of having a second parameter?

src/main/java/org/zalando/riptide/RestBase.java
whiskeysierra: Together with two javadoc comments that specify what the types are I'd say it's ok

whiskeysierra:
Do you need this in master? Otherwise I'd put this into release/2.0.0

src/main/java/org/zalando/riptide/RestBase.java
AlexanderYastrebov: no)

AlexanderYastrebov:
Up to you. This is simple, non breaking (hope so), using standard spring machinery

src/main/java/org/zalando/riptide/RestBase.java
AlexanderYastrebov: This is self describing without java doc. Also this is non-public api

whiskeysierra: It's not in line with Java Naming Conventions and it's not compliant to the current style, despite what the contribution guidelines state:

> keep the current code style

:stuck_out_tongue_winking_eye:

src/main/java/org/zalando/riptide/RestWithURL.java
```diff
@@ -0,0 +1,53 @@
...
+ * limitations under the License.
+ * ​⁣
+ */
+
+import java.net.URI;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+
+public class RestWithURL<D> {
```
whiskeysierra: final

src/test/java/org/zalando/riptide/WithUrlTest.java
```diff
@@ -0,0 +1,136 @@
...
+    @Test
+    public void shouldEncodeQueraparams() {
+        expectRequestTo("https://ru.wiktionary.org/w/index.php?title=%D0%A1%D0%BB%D1%83%D0%B6%D0%B5%D0%B1%D0%BD%D0%B0%D1%8F:%D0%9A%D0%BE%D0%BB%D0%BB%D0%B5%D0%BA%D1%86%D0%B8%D1%8F_%D0%BA%D0%BD%D0%B8%D0%B3&bookcmd=book_creator&referer=%D0%97%D0%B0%D0%B3%D0%BB%D0%B0%D0%B2%D0%BD%D0%B0%D1%8F%20%D1%81%D1%82%D1%80%D0%B0%D0%BD%D0%B8%D1%86%D0%B0");
+
+        unit.withUrl("https://ru.wiktionary.org/w/index.php?title={title}&bookcmd=book_creator&referer={referer}", "Служебная:Коллекция_книг", "Заглавная страница")
+            .execute(GET);
+    }
+
+    @Test
+    public void shouldMakeCoverageHappy() {
```
whiskeysierra: :confused:

src/main/java/org/zalando/riptide/RestBase.java
```diff
@@ -0,0 +1,68 @@
...
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.web.util.UriTemplateHandler;
+
+abstract class RestBase<T, D> {
+
+    protected final Router router = new Router();
+    protected final T template;
+
+    private final Supplier<UriTemplateHandler> uriTemplateHandler;
```
whiskeysierra: Why is this a supplier? Could we just use a single instance?

whiskeysierra:
> Up to you. This is simple, non breaking (hope so), using standard spring machinery

I'd like to have the current master freezed (in terms if features), i.e. any new feature should go to `release/2.0.0`. Can you rebase your changes and re-open a similar pull-request?

src/test/java/org/zalando/riptide/WithUrlTest.java
whiskeysierra: Can you make those into real tests?

AlexanderYastrebov: But its just a combinatoric explosion

src/main/java/org/zalando/riptide/RestBase.java
AlexanderYastrebov: I do not want to cache uriTemplateHandler provided some strange users change it in runtime for underlying RestTemlate

whiskeysierra:
Please rebase onto release/2.0.0

PR #81, opened 2016-05-20
Implemented support for URL template

AlexanderYastrebov:


tkrop:
:+1:

I stumbled over the use of the supplier pattern for the UriTemplateHandler, but I could not find a better solution. The problem is, that the support for URL templates spreads the use of the (Asynch)RestTemplate over two places. RestBase.execute(HttpMethod, URI, HttpEntity) and RestBase.withUrl(...). Therefore, I think it does not matter any longer whether RestBase receives a supplier or the original UriTemplateHandler in its constructor. In both cases the effect is, that a may surprised  that a change to the template is reflect/or not reflected when continuing the creation of the riptide Rest client.

May be the best solution is to freeze the RestTemplate by creating a defensive copy when creating a riptide Rest client to prevent subsequent changes having effect to the creation process.

tkrop:
:+1:

src/test/java/org/zalando/riptide/WithUrlTest.java
```diff
@@ -0,0 +1,145 @@
...
+        this.template = new RestTemplate();
+        this.server = MockRestServiceServer.createServer(template);
+        this.unit = Rest.create(template);
+    }
+
+    @Test
+    public void shouldExpandWithoutVariables() {
+        expectRequestTo("https://api.example.com/123");
+
+        unit.withUrl("https://api.example.com/123")
```
lukasniemeier-zalando: Looking at this I am not so sure about the naming `withUrl`. As Riptide is aiming for a fluently fluent API what about turning this around: `rest.execute(GET).on(url)`. Should also work fine with Post params.

AlexanderYastrebov: 1. I wanted to make it a nonbreaking change for current API
2. Are you also proposing `rest.execute(GET).headers().on(url)` for example?

AlexanderYastrebov: Also the purpose of this is to support templating. You can easily use Rest without it

whiskeysierra: > I wanted to make it a nonbreaking change for current API

That is still possible, since it would overload all execute methods without a URI as the second parameter.

whiskeysierra: > Are you also proposing rest.execute(GET).headers(headers).on(url) for example?

I'd only do this for the URI templating.

whiskeysierra:
Does URI templating has to be a part of the API like this? could it be something *external*:

```java
rest.execute(GET, expand("https://example.org/api/things/{id}", 123)).dispatch(status(),
    on(OK).call(..),
    ..);
```

src/main/java/org/zalando/riptide/RestBase.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+import java.net.URI;
+import java.util.Map;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.web.util.UriTemplateHandler;
+
+abstract class RestBase<D> {
```
whiskeysierra: If this class isn't public, but its methods are, does that make the methods effectively public in the two sub classes?

whiskeysierra: Side note: This would be a lot easier if our two different Dispatchers would share a common interface, but unfortunately the dispatch method uses a generic array and that is only safe (no compiler warnings) with a @SafeVarargs annotation, which is not possible to be placed on an interface. We could make it into an abstract class, maybe?

AlexanderYastrebov:
@whiskeysierra 
It is even possible to do tis:
`rest.execute(GET, template.getUriTemplateHandler().expand("https://example.org/api/things/{id}", 123))`
so this PR is not needed at all.
But the purpose is to have it on 'Rest' class so it is easy to use and discover with code completion.

src/main/java/org/zalando/riptide/RestBase.java
AlexanderYastrebov: > If this class isn't public, but its methods are, does that make the methods effectively public in the two sub classes?
Good question for interview)

AlexanderYastrebov:
@whiskeysierra 
Side note: In general for fluent API design I would prefer to minimize usage of static imports as this makes this API not explorable by code completion.

whiskeysierra:
> @whiskeysierra 
> It is even possible to do this:
> rest.execute(GET, template.getUriTemplateHandler().expand("https://example.org/api/things/{id}", 123))
> so this PR is not needed at all.

My sarcasm senses are tingling.

> But the purpose is to have it on Rest class so it is easy to use and discover with code completion.

:+1: 

> @whiskeysierra 
> Side note: In general for fluent API design I would prefer to minimize usage of static imports as this makes this API not explorable by code completion.

While on the other hand fluent API require to change the original code base, why static imports could come from even other libraries.

> E.g. having HttpMethod enum as argument is ok as code completion can suggest values, but using static method that produces general type like String or URI is not ok as there are tons of such methods and code completion cant suggest meaningful one

Agreed.

whiskeysierra:
:+1:

Issue #76, opened 2016-05-19
Update README


whiskeysierra:
- [x] add title, e.g. *Riptide: Client-side Response Routing*
- [x] add descriptions to selectors
- [x] add origin/reason for existence
- [x] document that the underlying RestTemplate shouldn't be changed afterwards
- [x] define Route and RoutingTree
- [x] Visualize routing tree
- [x] How to systematically produce a routing tree
- [x] How to programmatically modify the routing tree

lukasniemeier-zalando:
- [ ] remove all references to `RestTemplate`
- [ ] fix formatting in [stream doc](https://github.com/zalando/riptide/blob/release/2.0.0/riptide-stream/README.md)
- [ ] clarify last sentence of [stream doc](https://github.com/zalando/riptide/blob/release/2.0.0/riptide-stream/README.md)

Issue #67, opened 2016-05-17
Document how to systematically create routing trees


whiskeysierra:

tkrop:
I do not understand what you actually mean with that, otherwise I would contribute.

whiskeysierra:
The idea was to document how one should use riptide, i.e. how to build up a routing tree in code. I guess this is covered by #76

PR #82, opened 2016-05-20
Added test to visualize lambda support

### LINKED ISSUE #66: Allow lambdas to be used everywhere
right now we require method references or lambdas with type information for most of the capture/call methods

whiskeysierra:
Fixes #66

### DIFF
--- a/src/test/java/org/zalando/riptide/LambdaTest.java
+++ b/src/test/java/org/zalando/riptide/LambdaTest.java
@@ -1,0 +1,107 @@
+package org.zalando.riptide;
+
+/*
+ * ⁣​
+ * Riptide
+ * ⁣⁣
+ * Copyright (C) 2015 - 2016 Zalando SE
+ * ⁣⁣
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ​⁣
+ */
+
+import com.google.common.reflect.TypeToken;
+import org.junit.Test;
+import org.springframework.http.ResponseEntity;
+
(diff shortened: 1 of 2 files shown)


src/test/java/org/zalando/riptide/LambdaTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+import com.google.common.reflect.TypeToken;
+import org.junit.Test;
+import org.springframework.http.ResponseEntity;
+
+import static org.zalando.riptide.Conditions.anyStatus;
+import static org.zalando.riptide.Selectors.status;
+
+/**
+ * Tests whether all callbacks/functions can be supplied as lambdas. If this class compiles everything is already fine.
```
tkrop: Funny. Compile time testing ... didn't even need to annotate with \@Test.

But, do we really need to test this explicitly?

lukasniemeier-zalando: Users would be better off with a comprehensive Riptide example dispatch tree including some lambdas IMHO. I see value in this test class in a sense that future refactorings have to comply with it :wink:

whiskeysierra: > Users would be better off with a comprehensive Riptide example dispatch tree including some lambdas IMHO

Improving the readme for 2.0.0 is still an open issue, see #76 and #67

whiskeysierra: > I see value in this test class in a sense that future refactorings have to comply with it

Exactly, that was the idea, to have a regression test

whiskeysierra:
Anyway, nobody said something about my question. Is the required type hint for the overloaded versions of capture/call ok?

tkrop:
I can't see any difference in the behavior. Could you explain what I have missed?

whiskeysierra:
> I can't see any difference in the behavior. Could you explain what I have missed?

The question was whether this would be ok:

```java
on(200).call(Success.class, (Success sucess) -> {..});
```

Because this doesn't work (due to method overloading):
```java
on(200).call(Success.class, sucess -> {..});
```

tkrop:
Yes, but this does not work in the previous version either.

tkrop:
:+1:

whiskeysierra:
> Yes, but this does not work in the previous version either.

I know, I justed want to a) bring this to make this a conscious decision and b) to have a regression test that we don't loose the ability to supply lambdas in a future refactoring by accident.

whiskeysierra:
Why is zappr not eating the approval?

tkrop:
:+1:

Issue #85, opened 2016-05-23
Support reuse of routers for routing trees.


tkrop:
Currently, users users have to create the routing tree for each Rest instance. As far as I can see the routing tree for many request is pretty the same (pretty likely thread safe) and can be often declared static in advance of the request and reused on different occasions.

To allow declaration of independent routing trees, on solution would be support the stand alone creation and reuse of routers for routing trees. An example usage would look as follows:

    private static final Router<MediaType> CONTENT_ROUTER = create(contentType(),
            on(TEXT_PLAIN).capture(String.class),
            anyContentType().call(MyTest::failContent));

    private static final Router<HttpStatus> STATUS_ROUTER = create(status(),
            on(OK).dispatch(CONTENT_ROUTER),
            on(BAD_GATEWAY).call(MyTest::badGateway),
            anyStatus().call(MyTest::failStatus));

...

        Rest rest = Rest.create(new RestTemplate());
        rest.execute(GET, URI.create("http://my.uri/api"))
                .dispatch(STATUS_ROUTER)
                .as(String.class);

whiskeysierra:
Fixed with #86

Issue #73, opened 2016-05-19
Add possibility to include external routing trees


whiskeysierra:
```java
rest.execute(GET, url).dispatch(series(),
    on(SUCCESSFUL).capture(),
    anySeries().embed(handleProblems()));
```

Possible suggestions for naming:

- embed
- include
- route
- delegate
- delegateTo

whiskeysierra:
Fixed with #77

PR #86, opened 2016-05-24
Support reuse of routers for routing trees.

tkrop:
Proposed solution for #85

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,12 +365,11 @@ private Binding<HttpStatus> problems(final Condition<HttpStatus> condition) {
 }
 ```
 
-#### Template Routing
+#### Routing Templates
 
-Routing and response handling is sometimes pretty similar for many request, and can even be handle using the same
-defined routing logic for different requests. To enable reuse of this routing logic, Riptide supports the creation
-of nested *template routing*, that can be reused for dispatching different request. The following example defines
-a nested hierarchy of routing templates:
+Response handling is sometimes pretty similar for many types of requests, and can often be handled using the same
+routing logic. To enable reuse of routing logic, Riptide supports the creation of *routing templates*, that can be
+used for dispatching any number of responses. The following example defines a nested hierarchy of routing templates to support similar expected behaviors:
 
 ```java
 static final Router<MediaType> CONTENT_ROUTER = Router.create(contentType(),
@@ -391,8 +390,7 @@ static final Router<HttpStatus.Series> SERIES_ROUTER = Router.create(series(),
         anySeries().call(Example::failStatus));
 ```
 
-These templates can be used as follows to extend their functionality with a retry mechanism on internal server
-errors:
+These templates can be used and adapted as follows to support a retry mechanism on internal server errors:
 
 ```java
 Rest rest = Rest.create(new RestTemplate());
@@ -403,12 +401,9 @@ Partner partner = rest.execute(HttpMethod.GET, URI.create("http://example.com/ap
 
 ```
 
-Note: the need to change the routing template derives from the fact, that it needs to access an instance property.
-If this is not the case, the routing template can be used unchanged.
-
 Warning: be careful that all methods and instances referenced by the routing template are thread safe.
 
-### Patterns and examples
+### Patterns and Examples
 
 This section contains some ready to be used patterns and examples on how to solve certain challenges using Riptide: 
 
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ You are free to write your own, which requires you to implement the following in
 public interface Selector<A> {
 
     @Nullable
-    Binding<A> select(final ClientHttpResponse response, final Collection<Binding<A>> bindings) throws IOException;
+    Binding<A> select(final ClientHttpResponse response, final Map<A, Binding<A>> bindings) throws IOException;
 
 }
 ```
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -168,7 +168,7 @@ You are free to write your own, which requires you to implement the following in
 public interface Selector<A> {
 
     @Nullable
-    Binding<A> select(final ClientHttpResponse response, final Collection<Binding<A>> bindings) throws IOException;
+    Binding<A> select(final ClientHttpResponse response, final Map<A, Binding<A>> bindings) throws IOException;
```
whiskeysierra: Why does this have to be a map? I explicitly changed this to a Collection because it may not always make sense to think of the bindings as a map.

README.md
```diff
@@ -365,6 +365,49 @@ private Binding<HttpStatus> problems(final Condition<HttpStatus> condition) {
 }
 ```
 
+#### Template Routing
+
+Routing and response handling is sometimes pretty similar for many request, and can even be handle using the same
```
whiskeysierra: Riptide is handling responses, not requests.

README.md
```diff
@@ -365,6 +365,49 @@ private Binding<HttpStatus> problems(final Condition<HttpStatus> condition) {
 }
 ```
 
+#### Template Routing
```
whiskeysierra: I'd call this *Routing Templates*

README.md
```diff
@@ -365,6 +365,49 @@ private Binding<HttpStatus> problems(final Condition<HttpStatus> condition) {
...
+
+Routing and response handling is sometimes pretty similar for many request, and can even be handle using the same
+defined routing logic for different requests. To enable reuse of this routing logic, Riptide supports the creation
+of nested *template routing*, that can be reused for dispatching different request. The following example defines
+a nested hierarchy of routing templates:
+
+```java
+static final Router<MediaType> CONTENT_ROUTER = Router.create(contentType(),
```
whiskeysierra: The whole pull-request works on the assumption that it's expensive to re-create a routing tree for every invocation. Is that the case? Can we measure both versions? It pretty much sounds like premature optimization to me.

README.md
```diff
@@ -365,6 +365,49 @@ private Binding<HttpStatus> problems(final Condition<HttpStatus> condition) {
...
+        on(CONTRACT).capture(Contract.class),
+        on(PROBLEM).call(Problem.class, Example::propagate),
+        on(ERROR).call(Error.class, Example::propagate),
+        anyContentType().call(Example::failContent));
+
+static final Router<HttpStatus> STATUS_ROUTER = Router.create(status(),
+        on(OK).dispatch(CONTENT_ROUTER),
+        anyStatus().call(Example::failStatus));
+
+static final Router<HttpStatus.Series> SERIES_ROUTER = Router.create(series(),
```
whiskeysierra: Assuming that the gateway who holds those routes is a singleton they don't need to be static, but could be instance fields. That would solve the problem with the instance access as well.

src/main/java/org/zalando/riptide/Router.java
```diff
@@ -28,22 +28,55 @@
...
+public final class Router<A> {
 
-    final <A, S> Capture route(final ClientHttpResponse response, final List<HttpMessageConverter<?>> converters,
-            final Selector<A> selector, final Collection<Binding<A>> bindings) {
+    private final Selector<A> selector;
+    private final Map<A, Binding<A>> bindings;
 
-        @Nullable
-        final Binding<A> wildcard = findWildcard(bindings);
+    private Router(Selector<A> selector, Map<A, Binding<A>> bindings) {
```
whiskeysierra: final parameters

src/main/java/org/zalando/riptide/Router.java
```diff
@@ -28,22 +28,55 @@
...
+    @SafeVarargs
+    public static <A> Router<A> create(final Selector<A> selector, final Binding<A>... bindings) {
+        return create(selector, asList(bindings));
+    }
+
+    public static <A> Router<A> create(final Selector<A> selector, final Collection<Binding<A>> bindings) {
+        return new Router<A>(selector, Router.createMap(bindings));
+    }
+
+    private static <A> Map<A, Binding<A>> createMap(Collection<Binding<A>> bindings) {
```
whiskeysierra: I'd leave this up to the Selector implementation

src/main/java/org/zalando/riptide/Router.java
```diff
@@ -105,4 +122,11 @@
         }
     }
 
+    @SafeVarargs
+    public final Router<A> add(Binding<A>... bindings) {
+        Map<A, Binding<A>> map = Router.createMap(asList(bindings));
+        return new Router<A>(selector,
+                Stream.concat(this.bindings.entrySet().stream(), map.entrySet().stream())
```
whiskeysierra: This does not define how clashes are handled, does it? Can I override it?

whiskeysierra: > This does not define how clashes are handled, does it? Can I override it?

Ah, I just saw it.

whiskeysierra: Nevertheless, we should specify this somewhere that bindings are overwritten based on equality of the attribute.

src/test/java/org/zalando/riptide/RouterTest.java
```diff
@@ -47,26 +50,62 @@
...
+
+    @Test
+    public void shouldCreateNewRouterIfNotChanged() {
+        Router<HttpStatus> router = Router.create(status(), on(OK).capture());
+        Router<HttpStatus> result = router.add(on(OK).capture());
+        Assert.assertNotEquals(router, result);
+    }
+
+    @Test
+    public void shouldSubstituteBinding() {
```
whiskeysierra: Would be nice to have the same test for wildcards, since they are implemented as null I'd assume it could get tricky there.

README.md
tkrop: For bindings you are right, but for Routers and Selectors it is more natural thing to think of maps between attributes and and executors. It would be even possible to remove the Bindings here. In addition this is an internal change not visible in the API.

README.md
tkrop: Yes, I did this originally. I can change this back.

README.md
tkrop: > Riptide is handling responses, not requests.

This is stated in the begin of the sentence.

README.md
tkrop: No, it is not. It is stated on the idea, that definitions can be much smaller, if you once define the routing template and than change/reuse it for many different requests points.

README.md
tkrop: Agreed, but it is the simpler case to care for. I just wanted to give the context, that Router definitions are immutable and behave like singletons.

README.md
whiskeysierra: > No, it is not. It is stated on the idea, that definitions can be much smaller, if you once define the routing template and than change/reuse it for many different requests points.

Ok, but do we have the evidence that this is a common case? Do you intend to share routing trees across gateways? Or inside a gateway across different kind of HTTP requests to the same remote service?

I feel like externalizing the whole tree removes a lot of visibility. I have the same problem with constants in general, one always has to look them up while reading/understanding the code. Isn't embedding re-used parts (#73) enough for this?

src/main/java/org/zalando/riptide/Router.java
tkrop: > Nevertheless, we should specify this somewhere that bindings are overwritten based on equality of the attribute.

I thought about this - and nearly implemented it -, but than I realized that we need to include the Executors for proper function, but these will never be simple to compare - only on byte code level. Thus, I skipped this idea and decided that it would be acceptable to copy the router, even if no real change is done.

src/test/java/org/zalando/riptide/RouterTest.java
tkrop: > Would be nice to have the same test for wildcards, since they are implemented as null I'd assume it could get tricky there.

Not from code perspective, but you are right from behavior perspective.

src/main/java/org/zalando/riptide/Router.java
tkrop: > I'd leave this up to the Selector implementation

The maps are used and generated by the router. So I find this the natural private place.

README.md
tkrop: Oh, I remember: I renamed it in the end to be consistent with the naming of the previous sub-chapter.

README.md
tkrop: > Ok, but do we have the evidence that this is a common case?

At least I would do it in this way.

> Do you intend to share routing trees across gateways?

It would be possible if APIs of the gateways follow the same rules/patterns as suggested by our API guidelines.

> Or inside a gateway across different kind of HTTP requests to the same remote service?

Yes, I would do. For most APIs you pretty much expect the same behavior for each response. Even the differences between GET, PUT, and POST are mostly very small and can be handled by the same set of rules:

Status based routing is taken care of by the status router, content based routing is taken care of by the content based routing, etc.

whiskeysierra:
:+1:

Issue #102, opened 2016-05-31
Why always buffer the response?


jhorstmann:
In https://github.com/zalando/riptide/blob/master/src/main/java/org/zalando/riptide/Rest.java#L74 the response is always buffered. If I understand the history correctly this should only be needed in the `OAuth2CompatibilityResponseErrorHandler`.

When I replace the extractor `BufferingClientHttpResponse::buffer` with just `r -> r` all tests are still passing, so if this behaviour is really needed there should be a test documenting why it's needed.

whiskeysierra:
> When I replace the extractor BufferingClientHttpResponse::buffer with just r -> r all tests are still passing

Are we using a ByteArrayInputStream?

jhorstmann:
> Are we using a ByteArrayInputStream

Probably, but it's still an InputStream that can only be read once. The way I understand it, buffering is only needed for oauth errors and even there it [looks like our error handler would get an already buffered response][1].

 [1]: https://github.com/spring-projects/spring-security-oauth/blob/master/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/client/http/OAuth2ErrorHandler.java#L151

whiskeysierra:
I can't remember why we did it. @lukasniemeier-zalando Do you?

lukasniemeier-zalando:
I also don't remember why, my guess would have been to protect against error handlers (like OAuth). Or does the `RestTemplate` itself close the stream in `execute`?

jhorstmann:
Fixed by PR #104

PR #104, opened 2016-06-06
Do not buffer response unless needed

jhorstmann:
This makes it possible to extract the original response body, for example to handle streaming processing.

Related to question #102


jhorstmann:
It looks like the usecase of extracting an `InputStream` is not supported at all by springs `RestTemplate`. At the end of processing, the response is always closed:

https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/web/client/RestTemplate.java#L638

This works in unit tests with mock mvc because the `MockClientHttpResponse` ignores calls to the `close` method:

https://github.com/spring-projects/spring-test-mvc/blob/master/src/main/java/org/springframework/mock/http/client/MockClientHttpResponse.java#L67

jhorstmann:
@whiskeysierra What would you think about removing usage of `RestTemplate` in favour of using `ClientHttpRequestFactory` and `HttpMessageConverter` directly? Seems to be an easy change, the only problem would be the `OAuth2RestTemplate` implementations overriding `createRequest`.

lukasniemeier-zalando:
This looks very promising because it removes the need for the strange `ErrorHandler`s. Is the problem with the `AsyncRest` solved?

whiskeysierra:
> @whiskeysierra What would you think about removing usage of RestTemplate in favour of using ClientHttpRequestFactory and HttpMessageConverter directly?

We also need a `UriTemplateHandler` for the base URl + templating. Other than that I think it's fine. We don't really use the RestTemplate directly anyway, so this pretty much makes it more transparent to our clients what we actually require in order to work.

> This looks very promising because it removes the need for the strange ErrorHandlers.

:tada: 

> Is the problem with the AsyncRest solved?

What do you mean?

whiskeysierra:
Btw, all those API changes require that we move this PR to release/2.0.0

src/main/java/org/zalando/riptide/AsyncRest.java
```diff
@@ -59,17 +67,38 @@ public AsyncDispatcher execute(final HttpMethod method, final URI url, final Htt
...
+            final ExceptionWrappingFuture future = new ExceptionWrappingFuture(responseFuture);
+
+            return new AsyncDispatcher(converters, future, router);
+        } catch (IOException ex) {
+            final String message = String.format("I/O error on %s request for \"%s\":%s", method.name(), url, ex.getMessage());
+            throw new ResourceAccessException(message, ex);
+        }
+    }
+
+    static class ExceptionWrappingFuture extends ListenableFutureAdapter<ClientHttpResponse, ClientHttpResponse> {
```
jhorstmann: This is the weird async error handling that @lukasniemeier mentioned. Needs a better name but works.

jhorstmann:
> Btw, all those API changes require that we move this PR to release/2.0.0

I tried to keep the api compatible and only deprecated the no longer needed `OAuth2CompatibilityResponseErrorHandler` and `PassThroughResponseErrorHandler`, the other changes should all be internal.

The behaviour might be slightly different in case someone uses a `RestTemplate` with a custom error handler though.

src/main/java/org/zalando/riptide/RequestUtil.java
```diff
@@ -0,0 +1,79 @@
...
+import org.springframework.web.client.RestClientException;
+
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.util.List;
+import java.util.Optional;
+
+import static java.lang.String.format;
+
+class RequestUtil {
```
jhorstmann: Better name?

whiskeysierra: We only need this once, don't we?

src/test/java/org/springframework/mock/http/client/MockClientHttpResponse.java
```diff
@@ -0,0 +1,76 @@
...
+    @Override
+    public String getStatusText() throws IOException {
+        return this.status.getReasonPhrase();
+    }
+
+    @Override
+    public void close() {
+        try {
+            this.getBody().close();
+        } catch (IOException e) {
```
whiskeysierra: oO

src/test/java/org/zalando/riptide/AnyDispatchTest.java
```diff
@@ -52,8 +52,8 @@
     public AnyDispatchTest() {
         final RestTemplate template = new RestTemplate();
         template.setErrorHandler(new PassThroughResponseErrorHandler());
-        this.unit = Rest.create(template);
         this.server = MockRestServiceServer.createServer(template);
+        this.unit = Rest.create(template);
```
whiskeysierra: Very subtle. We should change the signature of `Rest.create(..)` to make this obvious.

src/main/java/org/zalando/riptide/Rest.java
```diff
@@ -69,16 +73,23 @@ public Dispatcher execute(final HttpMethod method, final URI url, final HttpHead
...
         }
     }
 
+    public static Rest create(final ClientHttpRequestFactory clientHttpRequestFactory, final List<HttpMessageConverter<?>> converters) {
+        return new Rest(clientHttpRequestFactory, converters);
+    }
+
     public static Rest create(final RestTemplate template) {
-        return new Rest(template);
+        return create(template.getRequestFactory(), template.getMessageConverters());
```
jhorstmann: Creating a `ClientHttpRequestFactory` that delegates to `template.createRequest()` might restore support for OAuth `Authorization` headers.

src/main/java/org/zalando/riptide/RequestUtil.java
whiskeysierra: I'd not put this into a static class if we only ever need this at a single place. A private instance method is good enough for me.

whiskeysierra:
> I tried to keep the api compatible and only deprecated the no longer needed OAuth2CompatibilityResponseErrorHandler and PassThroughResponseErrorHandler, the other changes should all be internal.
> 
> The behaviour might be slightly different in case someone uses a RestTemplate with a custom error handler though.

We inofficially went for a feature freeze for 1.x. I'd argue that those changes are significant enough that they should go into 2.0. I'm hoping we can get a RC in the next 1-2 weeks.

jhorstmann:
> We also need a `UriTemplateHandler` for the base URl + templating

I was looking for this but did not see any test cases. Or is this still only in PR #105 ?

whiskeysierra:
> I was looking for this but did not see any test cases. Or is this still only in PR #105 ?

No, it's in `release/2.0.0`

src/main/java/org/zalando/riptide/Rest.java
whiskeysierra: > Creating a ClientHttpRequestFactory that delegates to template.createRequest() might restore support for OAuth Authorization headers.

I think the whole support for `OAuth2RestTemplate` is questionable. If the version from stups (https://github.com/zalando-stups/stups-spring-oauth2-support/blob/master/stups-spring-oauth2-client/src/main/java/org/zalando/stups/oauth2/spring/client/StupsOAuth2RestTemplate.java) is doing the job, that what is the official one offering in addition?

src/test/java/org/springframework/mock/http/client/MockClientHttpResponse.java
jhorstmann: It's more or less the same that the real `HttpClientResponse` implementations are doing. I feel dirty for classpath-shadowing this class, but the behaviour is closer to the real world.

src/main/java/org/zalando/riptide/RequestUtil.java
jhorstmann: > We only need this once, don't we?

It's used by both `Rest` and `AsyncRest`. Could be moved to `RestClient` when we merge this with your PR.

jhorstmann:
> No, it's in release/2.0.0

Ugh, did not see that. Rebasing will be fun.

whiskeysierra:
> Rebasing will be fun.

:boom:

src/main/java/org/zalando/riptide/Rest.java
lukasniemeier-zalando: IIRC it offers refreshing the access token transparently. Also the chance that users want to use riptide with a `OAuth2RestTemplate` is higher than the knowledge of the `StupsOAuth2RestTemplate`.

src/test/java/org/zalando/riptide/AnyDispatchTest.java
lukasniemeier-zalando: Do you want to change the signature to expect `AsyncClientHttpRequestFactory` and `List<HttpMessageConverter<?>>`?

src/main/java/org/zalando/riptide/Rest.java
whiskeysierra: > IIRC it offers refreshing the access token transparently.

Which the tokens library is doing  for us anyway, right?

> Also the chance that users want to use riptide with a OAuth2RestTemplate is higher than the knowledge of the StupsOAuth2RestTemplate.

True. But we could make a strong argument for it in our README.

src/test/java/org/zalando/riptide/AnyDispatchTest.java
whiskeysierra: > Do you want to change the signature to expect AsyncClientHttpRequestFactory and List<HttpMessageConverter<?>>?

If we move away from (Async)RestTemplate to our real dependencies, then yes. (+ UriTemplateHandler)

src/main/java/org/zalando/riptide/Rest.java
lukasniemeier-zalando: > Which the tokens library is doing for us anyway, right?

> True. But we could make a strong argument for it in our README.

Yes but with a slightly different semantic (constant refreshing vs. on-demand). Also [this discussion](https://github.com/zalando/riptide/pull/104/files#r66149691) relates to this. If we drop `RestTemplate` from the API users will not be confused about using `RestTemplate` vs. `OAuth2RestTemplate` vs. `StupsOAuth2RestTemplate` and would solve it own their own.

src/main/java/org/zalando/riptide/AsyncRest.java
lukasniemeier-zalando: woot. How? :boom:

jhorstmann: > Ah... here... but how did this work before?

Previously this was done in `AsyncRestTemplate`: https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/web/client/AsyncRestTemplate.java#L506

jhorstmann:
> :boom:

Rebased. Is there a way to retarget this PR to the new release or do I have to open another PR which looses the discussions?

whiskeysierra:
> or do I have to open another PR which looses the discussions?

Yes, unfortunately.

PR #105, opened 2016-06-06
Unified syntax

### LINKED ISSUE #99: Unify syntax
```java
rest.xyz(GET, "/products/{id}", id)
        .accept(MediaTypes.PRODUCT, MediaTypes.PROBLEM)
        .header("Foo", "bar")
        .dispatch(series(),
            on(SUCCESSFUL).capture(Product.class),
            anySeries().call(fail()));
```

```java
rest.xyz(GET, "/products/{id}", id)
        .dispatch(series(),
            on(SUCCESSFUL).capture(Product.class),
            anySeries().call(fail()));
```

- [x] find good name for initial method

whiskeysierra:
Fixes #99


src/main/java/org/zalando/riptide/DefaultRoutingTree.java
```diff
@@ -0,0 +1,122 @@
...
+    private final Navigator<A> navigator;
+    private final Map<A, Route> routes;
+    private final Optional<Route> wildcard;
+
+    DefaultRoutingTree(final Navigator<A> navigator, final List<Binding<A>> bindings) {
+        this(navigator, map(bindings));
+    }
+
+    private DefaultRoutingTree(final Navigator<A> navigator, final Map<A, Route> routes) {
+        this(navigator, unmodifiableMap(routes), Optional.ofNullable(routes.remove(null)));
```
jhorstmann: Works, but looks a bit strange to first wrap into an unmodifiable collection and then remove items from the backing collection.

src/main/java/org/zalando/riptide/Navigator.java
```diff
@@ -22,26 +22,25 @@
...
-     * @return an optional binding match, if found
-     */
-    @Nullable
-    Executor select(final ClientHttpResponse response, final Map<A, Executor> routes) throws IOException;
+@FunctionalInterface
+public interface Navigator<A> {
+
+    Optional<Route> navigate(final ClientHttpResponse response, final RoutingTree<A> tree) throws IOException;
+
+    default List<Binding<A>> merge(final List<Binding<A>> present, final List<Binding<A>> additional) {
```
jhorstmann: Only used from `DefaultRoutingTree` as far as I can see, might make sense to move it as a private method over there.

whiskeysierra: It's part of the interface on purpose. Merging bindings/trees is potentially very dependent on the semantics of the navigator. First-write wins, last-write wins, overwrite, ...

PR #111, opened 2016-06-08
Replace usage of RestTemplate with own logic

jhorstmann:
Reusing the configuration of request factory, message converters and uri template handlers


src/main/java/org/zalando/riptide/OAuth2CompatibilityResponseErrorHandler.java
```diff
@@ -49,6 +47,7 @@
  * by throwing an exception containing the response. {@link Rest} catches this exception and continues with
  * the normal execution path.
  */
+@Deprecated
 public final class OAuth2CompatibilityResponseErrorHandler implements ResponseErrorHandler {
```
whiskeysierra: I'd drop it.

src/main/java/org/zalando/riptide/PassThroughResponseErrorHandler.java
```diff
@@ -23,6 +23,7 @@
 import org.springframework.http.client.ClientHttpResponse;
 import org.springframework.web.client.ResponseErrorHandler;
 
+@Deprecated
 public final class PassThroughResponseErrorHandler implements ResponseErrorHandler {
```
whiskeysierra: Delete

src/main/java/org/zalando/riptide/Rest.java
```diff
@@ -58,15 +77,26 @@ protected UriTemplateHandler getUriTemplateHandler() {
...
 
+    public static Rest create(final ClientHttpRequestFactory clientHttpRequestFactory, final List<HttpMessageConverter<?>> converters, final UriTemplateHandler uriTemplateHandler) {
+        return new Rest(clientHttpRequestFactory, converters, uriTemplateHandler);
+    }
+
+    public static Rest create(final ClientHttpRequestFactory clientHttpRequestFactory, final List<HttpMessageConverter<?>> converters) {
+        return create(clientHttpRequestFactory, converters, new DefaultUriTemplateHandler());
+    }
+
     public static Rest create(final RestTemplate template) {
```
whiskeysierra: Here I'd add a `@Deprecated`

src/test/java/org/springframework/mock/http/client/MockClientHttpResponse.java
```diff
@@ -0,0 +1,76 @@
...
+
+import org.springframework.http.HttpStatus;
+import org.springframework.http.client.ClientHttpResponse;
+import org.springframework.mock.http.MockHttpInputMessage;
+import org.springframework.util.Assert;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+public class MockClientHttpResponse extends MockHttpInputMessage implements ClientHttpResponse {
```
whiskeysierra:
```java
@Hack
@OhNoYouDidnt
```

src/main/java/org/zalando/riptide/Rest.java
whiskeysierra: In addition we should specify here that an subsequent change to the RestTemplate after this method returns will have no effect on the returned instance.

src/main/java/org/zalando/riptide/RestBase.java
```diff
@@ -77,4 +88,43 @@ public D execute(final HttpMethod method, final URI url, final HttpHeaders heade
...
+        request.getHeaders().putAll(headers);
+
+        @Nullable final T body = entity.getBody();
+
+        if (body != null) {
+
+            final Class<?> type = body.getClass();
+            @Nullable final MediaType contentType = headers.getContentType();
+
+            final Optional<HttpMessageConverter<T>> match = converters.stream()
```
whiskeysierra: Should we sort them by specificity first?

src/test/java/org/zalando/riptide/InputStreamTest.java
```diff
@@ -0,0 +1,214 @@
...
+            return inputStream.skip(n);
+        }
+
+        @Override
+        public synchronized int available() throws IOException {
+            checkClosed();
+            return inputStream.available();
+        }
+    }
+    private final URI url = URI.create("https://api.example.com/blobs/123");
```
whiskeysierra: New line

src/test/java/org/zalando/riptide/InputStreamTest.java
```diff
@@ -0,0 +1,214 @@
...
+        this.unit = Rest.create(template);
+    }
+
+    @Test
+    public void shouldAllowCloseOnce() throws IOException {
+        final InputStream content = new CloseOnceInputStream(new byte[] {'b', 'l', 'o', 'b'});
+        content.close();
+        try {
+            content.close();
+            fail("Should prevent multiple close calls");
```
whiskeysierra: I'd use `ExpectedException` for tests like this

whiskeysierra:
:boom: 

    [WARNING] Rule violated for class org.zalando.riptide.RestBase: lines covered ratio is 0.9, but expected minimum is 1.0

src/main/java/org/zalando/riptide/OAuth2CompatibilityResponseErrorHandler.java
lukasniemeier-zalando: +1 because 2.0 breaks either way.

src/main/java/org/zalando/riptide/Rest.java
lukasniemeier-zalando: Maybe best to pick something else than `create` (clarifies it also fur users of the legacy riptide).

src/test/java/org/springframework/mock/http/client/MockClientHttpResponse.java
lukasniemeier-zalando: You mean those annotations should be on the original class? :wink:

src/main/java/org/zalando/riptide/RestBase.java
jhorstmann: > Should we sort them by specificity first?

Is there such a thing? I thought the order in this list should be the order they are tried.

lukasniemeier-zalando:
We might also need to adapt the first paragraphs of the README.....................

whiskeysierra:
> We might also need to adapt the first paragraphs of the README.

Ignore the README for now. I'm going to change this once we wrap up 2.0

src/test/java/org/springframework/mock/http/client/MockClientHttpResponse.java
jhorstmann: > @Hack @OhNoYouDidnt

Not sure we need to shadow this class anymore, it was useful to verify the behaviour of `RestTemplate.doExecute`.

jhorstmann:
> [WARNING] Rule violated for class org.zalando.riptide.RestBase: lines covered ratio is 0.9, but expected minimum is 1.0

Fixed, during the rebase I accidentally kept overrides of the methods from `RestBase`

src/main/java/org/zalando/riptide/RestBase.java
```diff
@@ -77,4 +88,43 @@ public D execute(final HttpMethod method, final URI url, final HttpHeaders heade
 
     protected abstract <T> D execute(final HttpMethod method, final URI url, final HttpEntity<T> entity);
 
+    static <T> void writeRequestEntity(final HttpEntity<T> entity, final ClientHttpRequest request, List<HttpMessageConverter<?>> converters) throws IOException {
```
jhorstmann: One thing that `RestTemplate` currently does is setting accept headers based on the registered message converters. That logic would have to be reimplemented here.

src/main/java/org/zalando/riptide/RestBase.java
whiskeysierra: I checked and Spring is only doing this when reading responses, but not for requests:question:

src/main/java/org/zalando/riptide/RestBase.java
whiskeysierra: Personally I'd prefer to set this header manually in my riptide call, but that could be a good fallback.

whiskeysierra:
+1

src/main/java/org/zalando/riptide/RestBase.java
jhorstmann: > I checked and Spring is only doing this when reading responses, but not for requests

Do you have a Link? I only found https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/web/client/HttpMessageConverterExtractor.java#L87 which is just going through the list.

Only sorting I found was for accept headers in https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/web/client/RestTemplate.java#L744

whiskeysierra: I guess I mixed up the usage of the converters in MVC and RestTemplate.

jhorstmann:
Still need a place to close the response body. It seems this is not consistently handled by all message converters, some close the input stream themselves and some don't.

whiskeysierra:
:+1:

Issue #103, opened 2016-06-03
Verify that capturing lists works


whiskeysierra:
```java
rest.get("/api").dispatch(series(),
        on(SUCCESSFUL).capture(Body.class, this::map))
        .to(listOf(String.class));

private List<String> map(final Body body) {
    // ...
}
```

whiskeysierra:
Assuming that we're going to discourage the use of arrays as top-level structures in responses (see zalando/restful-api-guidelines/63) maybe it makes sense that Riptide has to first-class citizen support for this. Nevertheless, it should still be possible without too much hassle (e.g. without unsafe casts).

lukasniemeier-zalando:
I'd go with having the user explicitly specifying the input and output types of the mapping function

```java
rest.get("/api").dispatch(series(),
        on(SUCCESSFUL).capture(Body.class, listOf(String.class), this::map))
        .to(listOf(String.class));
```

whiskeysierra:
First things first, I agree. **But** I tried this on a local branch and it's horrible, it's just :hankey:. When we require it for everyone, regardless of types (generic or not) we're basically punishing everyone equally for Java's shortcomings in terms of type erasure. That feels just wrong. Using Riptide's unique syntax was meant to be fun, this just is no fun, no fun at all :rage:. But enough with the rant, I'd propose to go for the following principles:

1. Make it easy and fun to use in 95%+ of all use cases...
2. but still support the remaining 5%, with a slightly less-fun API, if needed
3. Make it as hard to shoot yourself in the foot as possible

I'd propose the following changes:

1. Every version of `capture` and `call` will be overloaded to have a version without any type information and once with full type information (two overloaded versions, once with Class and once with TypeToken)
2. The ones with type information behave as expected and look like in the previous example
3. The overloaded versions without type information use type inference
  1. We use https://github.com/jhalterman/typetools on the given lambda during *declaration* time
  2. If it fails we throw an exception early (at declaration time, not when we actually dispatch)

Thoughts?

lukasniemeier-zalando:
But isn't _typetools_ relying on type information of a concrete subtype? And those mapping-functions are just generic `EntityFunction`s right now. What do I miss?

whiskeysierra:
But they *are* actual subtypes of `EntityFunction`. It basically allows to do this:

```java
on(200).capture((String body) -> new Name(body))
```

Automatically uses `String.class` as a type hint for the message conversion and `Name.class` as the type hint for the Capture instance.

[...]

PR #113, opened 2016-06-12
Type inference for argument and return types of method references

jhorstmann:
Possible solution for issue #103 

Not integrated into the api yet because of PR #105 which refactored everything

### DIFF
--- a/src/main/java/org/zalando/riptide/TypeInference.java
+++ b/src/main/java/org/zalando/riptide/TypeInference.java
@@ -1,0 +1,157 @@
+package org.zalando.riptide;
+
+/*
+ * ⁣​
+ * Riptide
+ * ⁣⁣
+ * Copyright (C) 2015 - 2016 Zalando SE
+ * ⁣⁣
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


jhorstmann:
Some interesting discussion here: https://bugs.eclipse.org/bugs/show_bug.cgi?id=449063

Sounds like it is intentional by javac to not emit the generic signature in some cases. The eclipse compiler now has a compiler option to support this.

jhorstmann:
[More discussion and even a patch for OpenJDK][1] and an [answer by Brian Goetz][2]:

> There doesn't seem to be anything *wrong* with having a signature 
attribute on synthetic methods; the fact that we don't have them now is 
more an artifact of how erasure is implemented in the compiler than a 
deliberate decision to strip these methods of generic type information. 

 [1]: http://mail.openjdk.java.net/pipermail/compiler-dev/2015-January/009220.html
 [2]: http://mail.openjdk.java.net/pipermail/compiler-dev/2015-January/009268.html

src/main/java/org/zalando/riptide/TypeInference.java
```diff
@@ -0,0 +1,180 @@
...
+        if (function.getClass().isSynthetic()) {
+            return getFunctionInfoForMethodReference(function);
+        } else {
+            return getFunctionInfoForClass(function, ThrowingFunction.class);
+        }
+    }
+
+    static <T> FunctionInfo<T, Void> forConsumer(final ThrowingConsumer<T> function) {
+
+        if (function.getClass().isSynthetic()) {
```
jhorstmann: I need to create a synthetic class with custom `writeReplace` method to test code coverage here. I have no idea how to do that. Even with creating a class from bytecode, the flag seems to be cleared on class loading.

whiskeysierra:
As discussed, we're going for a simpler API that doesn't need this.

PR #115, opened 2016-06-24
Support streaming responses

### LINKED ISSUE #101: Support streaming processing of responses
Not sure how well this would fit in with riptide and resttemplate, but it would be nice to have riptide support streaming of for example newline delimited json, as used in nakadi or the twitter api

jhorstmann:
Fixes #101 

Code coverage is currently failing because of try-with-resources which is not completely supported by jacoco.

### DIFF
--- a/src/main/java/org/zalando/riptide/Route.java
+++ b/src/main/java/org/zalando/riptide/Route.java
@@ -23,12 +23,22 @@ package org.zalando.riptide;
 import com.google.common.reflect.TypeToken;
 import org.springframework.http.client.ClientHttpResponse;
 
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
 import static org.zalando.riptide.Capture.none;
 
 @FunctionalInterface
 public interface Route {
 
     Capture execute(final ClientHttpResponse response, final MessageReader reader) throws Exception;
+
+    static <T> Route stream(final TypeToken<T> typeToken) {
+        return (response, reader) -> {
+            final Stream<T> stream = StreamSupport.stream(new ResponseSpliterator<>(response, reader, typeToken), false).onClose(response::close);
+            return Capture.ofStream(stream);
+        };
+    }
 
     static Route capture() {
         return (response, reader) ->
--- a/src/main/java/org/zalando/riptide/Route.java
+++ b/src/main/java/org/zalando/riptide/Route.java
@@ -52,15 +52,19 @@ public interface Route {
 
     static Route call(final ThrowingRunnable consumer) {
         return (response, reader) -> {
-            consumer.run();
-            return none();
+            try (final ClientHttpResponse responseToClose = response) {
+                consumer.run();
+                return none();
+            }
         };
     }
 
     static Route call(final ThrowingConsumer<ClientHttpResponse> consumer) {
         return (response, reader) -> {
-            consumer.accept(response);
-            return none();
+            try (final ClientHttpResponse responseToClose = response) {
+                consumer.accept(responseToClose);
+                return none();
+            }
         };
     }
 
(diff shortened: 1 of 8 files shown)


src/main/java/org/zalando/riptide/Route.java
```diff
@@ -42,15 +52,19 @@ static Route capture(final ThrowingFunction<ClientHttpResponse, ?> function) {
 
     static Route call(final ThrowingRunnable consumer) {
         return (response, reader) -> {
-            consumer.run();
-            return none();
+            try (final ClientHttpResponse responseToClose = response) {
```
whiskeysierra: That looks hacky.

src/test/java/org/zalando/riptide/CloseOnceInputStream.java
```diff
@@ -0,0 +1,102 @@
...
+ * limitations under the License.
+ * ​⁣
+ */
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Created by jhorstmann on 24.06.16.
```
whiskeysierra: Meh

src/test/java/org/zalando/riptide/StreamTest.java
```diff
@@ -0,0 +1,126 @@
...
+import static org.springframework.test.web.client.match.MockRestRequestMatchers.requestTo;
+import static org.springframework.test.web.client.response.MockRestResponseCreators.withSuccess;
+import static org.zalando.riptide.Bindings.on;
+import static org.zalando.riptide.Selectors.contentType;
+
+public final class StreamTest {
+
+    static class StreamingMappingJackson2HttpMessageConverter extends MappingJackson2HttpMessageConverter {
+
+        public static final MediaType APPLICATION_X_JSON_STREAM = new MediaType("application", "x-json-stream");
```
whiskeysierra: I think this one should produce a Stream<T> and it could be it's own project. It should also support application/json-seq

jhorstmann: > I think this one should produce a Stream and it could be it's own project.

That could also work

whiskeysierra:
> Code coverage is currently failing because of try-with-resources which is not completely supported by jacoco.

Can you replace try-with-resources with traditional try-finally blocks?

jhorstmann:
> Can you replace try-with-resources with traditional try-finally blocks?

I will try, probably the uncovered code is that the close method could also throw an exceptions and this would then be added to the supressed exceptions.

I will also investigate if this can be solved by a message converter alone. You could still have a look whether the refactoring to the `Capture` class makes sense.

whiskeysierra:
> I will also investigate if this can be solved by a message converter alone. You could still have a look whether the refactoring to the Capture class makes sense.

I think Capture should stay as it it. The majority of the streaming support should be in custom message covnerters, imho. The API changes in Riptide should be kept to a minimum, imo, e.g. only custom Actions:

```java
http.get("/").dispatch(statusCode(),
    on(200).call(stream(Event.class, event -> {})));
```

```java
http.get("/").dispatch(statusCode(),
    on(200).capture(streamOf(Event.class))
    .get().cast(streamOf(Event.class)).forEach(event -> {});
```

src/main/java/org/zalando/riptide/Route.java
tkrop: Why, its a language feature?

tkrop:
I agree implementation in this kind is ugly.

Regarding coverage: as I remember, it is the problem of jacoco that it cannot capture correctly on exceptions - finally included. Switching between two idioms will not change the result, as their byte code is similar (I think it is the same).

Anyhow, the routing is only conducted once, thus the reading loop should be handled by the called function.

@whiskeysierra your first above API example suggests to add a StreamConsumer function, that exposes the MessageReader and to provide a default implementation that extracts events in a loop and forwards them to a (Response)EntityFunction as the only changes. I like that idea.

For the second, I would suggest the following concise form, but currently cannot see the additional value of it:
```
http.get("/").dispatch(statusCode(),
    on(200).stream(Event.class))
    .get().stream(Event.class).forEach(event -> {});
```

I'm just wandering whether streams will always have same object types or whether they may be different.

whiskeysierra:
@tkrop Is working on an alternative.

Issue #106, opened 2016-06-07
Javadocs


whiskeysierra:
Add meaningful Javadocs to relevant parts of the public API

LappleApple:
Hey @whiskeysierra, anything more to do here? Do you think adding a "Help Wanted" label is appropriate?

PR #116, opened 2016-06-29
Async by default

### LINKED ISSUE #103: Verify that capturing lists works

### LINKED ISSUE #107: Check for consistent exception handling
- [ ] Are we catching/rethrowing/wrapping exceptions somewhere?
- [ ] If yes, should we?
- [ ] Define clear rules + tests for exception handling inside riptide itself
- [ ] Document it

### LINKED ISSUE #109: Make Riptide async by default?
Right now we always have to maintain two APIs: Rest (sync) and AsyncRest (async). Because of Riptide's design they both support the very same routing trees out of the box. This already makes Riptide stand out from the competition, since their APIs usually differ dramatically between the two approaches.

The problem we have right now is the different return value for async: `ListenableFuture<Capture>` vs. `Capture`. Translating from async to sync is relative easy:

```java
http.get(..)
        .dispatch(..)
        .to(Success.class);
```

vs.

```java
http.get(..)
        .dispatch(..)
        .get(10, SECONDS)
        .to(Success.class);
```

Given this, should we consider dropping the dedicated support for the synchronous version altogether?

whiskeysierra:
Fixes #109 
Fixes #107 
Fixes #103


src/main/java/org/zalando/riptide/RestClient.java
```diff
@@ -20,133 +20,79 @@
...
-import org.springframework.web.client.RestClientException;
 
-import javax.annotation.Nullable;
-import java.io.IOException;
 import java.net.URI;
-import java.util.List;
-import java.util.Optional;
 
-import static java.lang.String.format;
+abstract class RestClient {
```
lukasniemeier-zalando: Is this class still needed?

src/main/java/org/zalando/riptide/RestClient.java
```diff
@@ -20,133 +20,79 @@
...
 
-abstract class RestClient<R> {
-
-    public final Requester<R> get(final String urlTemplate, final Object... urlVariables) {
+    public final Requester get(final String urlTemplate, final Object... urlVariables) {
         return execute(HttpMethod.GET, urlTemplate, urlVariables);
     }
 
-    public final Requester<R> get(final URI url) {
+    public final Requester get(final URI url) {
```
lukasniemeier-zalando: IMO all those methods should be covered in #106 - clarifying the usage of the `UriTemplateHandler`

src/main/java/org/zalando/riptide/RestClient.java
whiskeysierra: No, it's not. Good catch.

tkrop:
:+1:

src/main/java/org/zalando/riptide/Rest.java
```diff
@@ -38,67 +43,205 @@
...
 
-        };
+            final Class<?> type = body.getClass();
+            @Nullable final MediaType contentType = headers.getContentType();
+
+            converters.stream()
+                    .filter(c -> c.canWrite(type, contentType))
+                    .map(this::<T>cast)
+                    .findFirst()
+                    .orElseThrow(() -> {
```
lukasniemeier-zalando: extract to method?

lukasniemeier-zalando:
:+1:

PR #119, opened 2016-07-05
Capture support

### LINKED ISSUE #117: Don't use Capture as a return value
By returning Capture (or future thereof) we have the burden to preserve type safety. If the change this we can let the client handle this in a safer and easier way:

```java
Capture<List<String>> capture = Capture.empty();

http.get("/").dispatch(series(),
        on(SUCCESSFUL).call(capture),
        anySeries().call(fail))
        .get(10, SECONDS);

capture.retrieve(); // throws NotReadyException vs EmptyException
```

**Benefits**
- completely type safe
- no unsafe casts
- easy to understand for clients
- API is now 100% callbacks
- does not even need to be part of the core API (even though it probably will)

This will complete get rid of the problematic of #114.

- [x] externalize into own module: `riptide-capture`

whiskeysierra:
Fixes #117 

- [ ] decide what to do with the one new *TODO*


riptide-capture/src/main/java/org/zalando/riptide/capture/DefaultCapture.java
```diff
@@ -0,0 +1,63 @@
...
+        if (value == null) {
+            throw new NoSuchElementException("No value present");
+        }
+
+        return value;
+    }
+
+    @Override
+    public ListenableFuture<T> adapt(final ListenableFuture<?> future) {
+        @SuppressWarnings("unchecked") // TODO fix
```
lukasniemeier-zalando: this is the only *new* *TODO* I have found

whiskeysierra: > this is the only new TODO I have found

I can't tell, is that appreciation, unbelief or disappointment?

lukasniemeier-zalando: (1) and (2). And the only thing I was able to contribute on mobile for now (90 files changed seems to be the cap for my device :wink:)

whiskeysierra: > 90 files changed seems to be the cap for my device

Most of them are just moved.

lukasniemeier-zalando:
:+1:

whiskeysierra:
The only way I see to get rid of the TODO (apart from just deleting it) is to do the adapter thingy ourselves.

PR #129, opened 2016-07-11
Feature streams

tkrop:


pom.xml
```diff
@@ -59,6 +59,7 @@
 
     <modules>
         <module>riptide-core</module>
+        <module>riptide-stream</module>
```
whiskeysierra: Not lexicographically ordered /nitpick

riptide-stream/README.md
```diff
@@ -0,0 +1,26 @@
...
+        rest.get("/repos/{org}/{repo}/contributors", "zalando", "riptide")
+                .accept(MediaType.APPLICATION_JSON)
+                .dispatch(series(),
+                       on(SUCCESSFUL).call(streamOf(User.class),
+                               forEach(user -> println(user.login + " (" + user.contributions + ")"))))
+                .get();
+    }
+}
+
+The unique entry point for all specific methods is the (Streams)[src/main/java/org/zalando/riptide/stream/Streams.java] class. *Note:* The stream converter is an replacement to the default spring JSON converter that does not support streaming, and thus should be not registered together with it.
```
whiskeysierra: Links are done the other way around, parenthesis first and square brackets later.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamConverter.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Override
+    public boolean canRead(Class<?> clazz, MediaType mediaType) {
+        return this.canRead(clazz, null, mediaType);
+    }
+
+    @Override
+    public boolean canRead(Type type, Class<?> contextClass, MediaType mediaType) {
+        final JavaType javaType = getJavaType(type, contextClass);
+        if (Stream.class.isAssignableFrom(javaType.getRawClass())) {
```
whiskeysierra: Why not just:

```java
return Stream.class.isAssignableFrom(javaType.getRawClass())
        && super.canRead(javaType.containedType(0), contextClass, mediaType);
```

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamConverter.java
```diff
@@ -0,0 +1,79 @@
...
+
+        JavaType javaType = getJavaType(type, contextClass);
+        try {
+            if (Stream.class.isAssignableFrom(javaType.getRawClass())) {
+                final JavaType containedType = javaType.containedType(0);
+                final InputStream stream = new StreamFilter(inputMessage.getBody());
+                final JsonParser parser = objectMapper.getFactory().createParser(stream);
+                return StreamSupport.stream(new StreamSpliterator<>(containedType, parser, objectMapper), false);
+            }
+            return objectMapper.readValue(inputMessage.getBody(), javaType);
```
whiskeysierra: Or is this used to support array like structures?

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamFilter.java
```diff
@@ -0,0 +1,93 @@
...
+            }
+        }
+        for (int index = last; index < until; index++) {
+            b[index] = 0;
+        }
+        return last - off;
+    }
+
+    @Override
+    public long skip(long n) throws IOException {
```
whiskeysierra: Why is this needed?

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamSpliterator.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    public boolean tryAdvance(Consumer<? super T> action) {
+        try {
+            JsonToken token = parser.nextToken();
+            if (token == null) {
+                return false;
+            }
+            if (!type.isArrayType() && !type.isCollectionLikeType()) {
```
whiskeysierra: I usually prefer non-negated cases first, i.e. `if (type.isArrayType() || type.isCollectionLikeType()) {`

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamSpliterator.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    public long estimateSize() {
+        return Long.MAX_VALUE;
+    }
+
+    @Override
+    public int characteristics() {
+        return ORDERED | IMMUTABLE;
```
whiskeysierra: I have no clue why Oracle thought it's a good idea to still use bitsets in 2016.

riptide-stream/src/main/java/org/zalando/riptide/stream/Streams.java
```diff
@@ -0,0 +1,114 @@
...
+import com.google.common.reflect.TypeParameter;
+import com.google.common.reflect.TypeToken;
+
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+
+/**
+ * Main entry point for <b>Riptide Streams</b> extension to capture arbitrary infinite object streams. It allows to
+ * receive infinite streams using application/x-json-stream and application/json-seq format, as well as simple finite
+ * streams from lists and arrays. Must be enable by registering the {@link StreamConverter} with Riptide (using
```
whiskeysierra: enable**d**

riptide-stream/src/main/java/org/zalando/riptide/stream/Streams.java
```diff
@@ -0,0 +1,114 @@
...
+ *     rest.get("/repos/{org}/{repo}/contributors", "zalando", "riptide")
+ *             .accept(MediaType.APPLICATION_JSON)
+ *             .dispatch(series(),
+ *                     on(SUCCESSFUL).call(streamOf(User.class),
+ *                             forEach(user -> println(user.login + " (" + user.contributions + ")"))))
+ *             .get();
+ * }
+ * </pre>
+ * 
+ * <b>Note:</b> The stream converter is an replacement to the default spring JSON converter that does not support
```
whiskeysierra: Why shouldn't it? They should be mutually exclusive.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamConverter.java
```diff
@@ -0,0 +1,79 @@
...
+import org.springframework.http.HttpInputMessage;
+import org.springframework.http.MediaType;
+import org.springframework.http.converter.HttpMessageNotReadableException;
+import org.springframework.http.converter.json.MappingJackson2HttpMessageConverter;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+class StreamConverter extends MappingJackson2HttpMessageConverter {
```
whiskeysierra: Extending is probably not the best way, since that exposes a lot of methods that we probably shouldn't expose. It's also using incompatible media types.

riptide-stream/src/main/java/org/zalando/riptide/stream/Streams.java
```diff
@@ -0,0 +1,114 @@
...
+ */
+public final class Streams {
+
+    public static <T> TypeToken<Stream<T>> streamOf(final Class<T> type) {
+        return streamOf(TypeToken.of(type));
+    }
+
+    @SuppressWarnings("serial")
+    public static <T> TypeToken<Stream<T>> streamOf(final TypeToken<T> type) {
+        final TypeToken<Stream<T>> listType = new TypeToken<Stream<T>>() {
```
whiskeysierra: streamType

riptide-stream/src/main/java/org/zalando/riptide/stream/Streams.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @SuppressWarnings("serial")
+    private static class UncheckedConsumerException extends RuntimeException {
+
+        public UncheckedConsumerException(Exception cause) {
+            super(cause);
+        }
+
+        @Override
+        public Exception getCause() {
```
whiskeysierra: Why is this needed?

whiskeysierra: Ah, I see now.

riptide-stream/src/main/java/org/zalando/riptide/stream/Streams.java
```diff
@@ -0,0 +1,114 @@
...
+                    return;
+                }
+                input.forEach(wrap(consumer));
+            } catch (UncheckedConsumerException ex) {
+                throw ex.getCause();
+            }
+        };
+    }
+
+    private static <I> Consumer<? super I> wrap(final ThrowingConsumer<I> consumer) throws Exception {
```
whiskeysierra: I think it would be clearer if this method would be inlined.

riptide-stream/src/test/java/org/zalando/riptide/stream/EnforceCoverageTest.java
```diff
@@ -0,0 +1,54 @@
...
+public final class EnforceCoverageTest {
+
+    @Test
+    public void shouldUseStreamsConstructor() {
+        new Streams();
+    }
+
+    @Test
+    public void shouldUseStreamConverterConstructor() {
+        Streams.streamConverter();
```
whiskeysierra: This should be tested in a real test, not here.

riptide-stream/src/test/java/org/zalando/riptide/stream/EnforceCoverageTest.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Test
+    public void shouldUseStreamSpliteratorTrySplit() {
+        assertNull(new StreamSpliterator<>(null, null, null).trySplit());
```
whiskeysierra: When would this officially be used? Parallel stream processing?

riptide-stream/src/test/java/org/zalando/riptide/stream/StreamConverterTest.java
```diff
@@ -0,0 +1,126 @@
...
+        Assert.assertTrue(unit.canRead(AccountBody[].class, null));
+    }
+
+    @Test
+    public void shouldReadStreams() throws Exception {
+        final StreamConverter unit = new StreamConverter();
+        Assert.assertTrue(unit.canRead(Streams.streamOf(List.class).getType(), null, null));
+        Assert.assertTrue(unit.canRead(Streams.streamOf(List[].class).getType(), null, null));
+        Assert.assertTrue(unit.canRead(Streams.streamOf(AccountBody.class).getType(), null, null));
+        Assert.assertTrue(unit.canRead(Streams.streamOf(AccountBody[].class).getType(), null, null));
```
whiskeysierra: There is no single negative test for canRead, so returning true would still make all tests pass.

riptide-stream/src/test/java/org/zalando/riptide/stream/StreamConverterTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class StreamConverterTest {
+
+    @Rule
+    public final ExpectedException exception = ExpectedException.none();
+
+    @Test
+    public void shouldRead() throws Exception {
```
whiskeysierra: It's missing real content types.

riptide-stream/src/test/java/org/zalando/riptide/stream/StreamConverterTest.java
```diff
@@ -0,0 +1,126 @@
...
+        Assert.assertTrue(unit.canRead(Streams.streamOf(List[].class).getType(), null, null));
+        Assert.assertTrue(unit.canRead(Streams.streamOf(AccountBody.class).getType(), null, null));
+        Assert.assertTrue(unit.canRead(Streams.streamOf(AccountBody[].class).getType(), null, null));
+    }
+
+    @Test
+    public void shouldReadCreateStream() throws Exception {
+        Type type = Streams.streamOf(AccountBody.class).getType();
+        final StreamConverter unit = new StreamConverter(new ObjectMapper().findAndRegisterModules());
+        HttpInputMessage input = new HttpInputMessage() {
```
whiskeysierra: Would probably be smaller by using Mockito (or a private method).

riptide-stream/src/test/java/org/zalando/riptide/stream/StreamsTest.java
```diff
@@ -0,0 +1,296 @@
...
+
+    public StreamsTest() {
+        final ObjectMapper mapper = new ObjectMapper().findAndRegisterModules();
+        MockSetup setup = new MockSetup(baseUrl, Arrays.asList(Streams.streamConverter(mapper)));
+        this.server = setup.getServer();
+        this.unit = setup.getRest();
+    }
+
+    @Test
+    @SuppressWarnings("unchecked")
```
whiskeysierra: This should be placed on the statement, not the method.

riptide-stream/src/test/java/org/zalando/riptide/stream/StreamsTest.java
```diff
@@ -0,0 +1,296 @@
...
+    public void shouldCallConsumerWithList() throws Exception {
+        final InputStream stream = new ClassPathResource("account-list.json").getInputStream();
+    
+        server.expect(requestTo(url)).andRespond(
+                withSuccess()
+                        .body(new InputStreamResource(stream))
```
whiskeysierra: `.body(new ClassPathResource(".."))`

riptide-stream/src/test/java/org/zalando/riptide/stream/StreamsTest.java
```diff
@@ -0,0 +1,296 @@
...
+        unit.get("/accounts").dispatch(status(),
+                on(OK).call(streamOf(listOf(AccountBody.class)), forEach(verifier)),
+                anyStatus().call(this::fail)).get();
+    
+        verify(verifier).accept(Arrays.asList(
+                new AccountBody("1234567890", "Acme Corporation"),
+                new AccountBody("1234567891", "Acme Company"),
+                new AccountBody("1234567892", "Acme GmbH"),
+                new AccountBody("1234567893", "Acme SE")));
+        verify(verifier, times(1)).accept(any(List.class));
```
whiskeysierra: times(1) is the default

riptide-stream/src/test/java/org/zalando/riptide/stream/StreamsTest.java
```diff
@@ -0,0 +1,296 @@
...
+        unit.get("/accounts").dispatch(status(),
+                on(OK).call(streamOf(AccountBody[].class), forEach(verifier)),
+                anyStatus().call(this::fail)).get();
+    
+        verify(verifier).accept(new AccountBody[] {
+                new AccountBody("1234567890", "Acme Corporation"),
+                new AccountBody("1234567891", "Acme Company"),
+                new AccountBody("1234567892", "Acme GmbH"),
+                new AccountBody("1234567893", "Acme SE") });
+        verify(verifier, times(1)).accept(any(AccountBody[].class));
```
whiskeysierra: Since Java 8 the any method doesn't need to get a parameter, type inference is now powerful enough.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamConverter.java
tkrop: Because of the second return case: It differs in javaType.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamConverter.java
tkrop: No,  this is the fallback to support all standard cases - the once without streaming.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamFilter.java
tkrop: Actually, to prevent failures when Jackson decides to use skip, which it currently does not.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamSpliterator.java
tkrop: Currently, there is no non-negated case. I can change it to `if (!(type.isArrayType() || type.isCollectionLikeType())) {`, if you like.

riptide-stream/src/main/java/org/zalando/riptide/stream/Streams.java
tkrop: Two nested lambdas in my opinion is to hard to read.

riptide-stream/src/test/java/org/zalando/riptide/stream/EnforceCoverageTest.java
tkrop: Would be nice, but the use case is in the sample service, which does not help with the coverage. I didn't like the idea to add it artificially to the other test cases.

riptide-stream/src/test/java/org/zalando/riptide/stream/EnforceCoverageTest.java
tkrop: Yes, as far as I understand.

riptide-stream/src/test/java/org/zalando/riptide/stream/StreamConverterTest.java
tkrop: Can you give an example? I thought `AccountBody` represents real content types.

riptide-stream/src/test/java/org/zalando/riptide/stream/StreamConverterTest.java
tkrop: Yes, I originally had one, but I fixed it that it now reads all, so I think there is none left - except for Media-Type-based once. I will add one of these.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamFilter.java
```diff
@@ -0,0 +1,93 @@
...
+        this(in, 8192);
+    }
+
+    protected StreamFilter(InputStream in, int size) {
+        super(in);
+        this.size = size;
+    }
+
+    private boolean filtered(byte read) {
+        return read == 30;
```
jhorstmann: Please add a constant clarifying that this is the ASCII record separator

riptide-stream/src/test/java/org/zalando/riptide/stream/StreamConverterTest.java
tkrop: Agreed. Mockito is a good option.

riptide-stream/src/main/java/org/zalando/riptide/stream/Streams.java
```diff
@@ -0,0 +1,114 @@
...
+ *                     on(SUCCESSFUL).call(streamOf(User.class),
+ *                             forEach(user -> println(user.login + " (" + user.contributions + ")"))))
+ *             .get();
+ * }
+ * </pre>
+ * 
+ * <b>Note:</b> The stream converter is an replacement to the default spring JSON converter that does not support
+ * streaming, and thus should be not registered together with it.
+ */
+public final class Streams {
```
jhorstmann: Could we find a better name that avoid confusion with `java.util.stream.Streams`?

riptide-stream/src/test/java/org/zalando/riptide/stream/StreamsTest.java
tkrop: You are right. Looks different in the code, but should work.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamSpliterator.java
jhorstmann: Will this work if I have an array nested inside the stream elements? The opening and closing tokens should only be consumed on the outermost level.

tkrop: Yes, they are. The splitter works only on the outermost element.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamConverter.java
whiskeysierra: Your **Stream**Converter shouldn't handle other cases, imo.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamFilter.java
whiskeysierra: How does this differ from the default implementation?

riptide-stream/src/main/java/org/zalando/riptide/stream/Streams.java
whiskeysierra: Streaming?

riptide-stream/src/main/java/org/zalando/riptide/stream/Streams.java
whiskeysierra: But it's easy to see where the UncheckedConsumerException is thrown/catched and why.

riptide-stream/src/test/java/org/zalando/riptide/stream/EnforceCoverageTest.java
whiskeysierra: If it's used by your SampleService it should be covered by a test to make sure it actually works as advertised.

riptide-stream/src/test/java/org/zalando/riptide/stream/StreamConverterTest.java
whiskeysierra: `application/json-seq` e.g. Your converter should only handle these , imo.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamConverter.java
whiskeysierra: My ideas to improve this class:
- don't extend MappingJackson2HttpMessageConverter, if you need it, use an instance field (maybe static?)
- don't support List/Array directly, there is no need since clients can just use MappingJackson2HttpMessageConverter and immediately call `List.stream()` on the return value
- only support two concrete media types: `application/json-seq` and `application/x-json-stream`
- don't try to compete with MappingJackson2HttpMessageConverter, but work in conjunction

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamConverter.java
tkrop: Maybe we should name it StreamCapableConverter?

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamFilter.java
tkrop: The default implementation returns the byte number including the filtered chars. If I do not implement filtering on skip, the expected skipped byte number will differ from the read byte number.

riptide-stream/src/main/java/org/zalando/riptide/stream/Streams.java
tkrop: Because the original converter catches Stream<?> also, but will only handle the first result and ignore the rest of the stream. The result of registering both depends on the order of registration. As it is hard to argue, which one should be loaded first for proper function, I think it would be use as an substitution of the original one. That's although the reason, why I made the converter capable as a full replacement instead of specializing it.

riptide-stream/src/main/java/org/zalando/riptide/stream/Streams.java
tkrop: Possible, was my first choice before I switched to the shorter version.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamConverter.java
tkrop: @whiskeysierra I'm still thinking about this. I liked the unified feature to support immediate streams for long but finite lists/arrays. The distinction between finite and infinite is sometimes blurring. Restricting the handler to the above media types would prevent this finally.

I'm with you that aggregation has to be preferred over inheritance, so I will adapt to this solution, and look how it will work out.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamFilter.java
whiskeysierra: But the default implementation delegates to your implementation of `read`, so it should do the right thing already.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamConverter.java
```diff
@@ -0,0 +1,166 @@
...
+import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.type.TypeFactory;
+
+class StreamConverter<T> implements GenericHttpMessageConverter<T> {
+
+    private static final List<MediaType> SUPPORTED_MEDIA_TYPES =
+            Arrays.asList(APPLICATION_JSON_SEQ,
+                    APPLICATION_X_JSON_STREAM,
+                    MediaType.APPLICATION_JSON_UTF8,
```
whiskeysierra: That makes it difficult to use it together with the MappingJackson2HttpMessageConverter, because both match the same media types.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamConverter.java
```diff
@@ -0,0 +1,166 @@
...
+    private final ObjectMapper mapper;
+    private final List<MediaType> medias;
+
+    public StreamConverter() {
+        this(Jackson2ObjectMapperBuilder.json().build());
+    }
+
+    public StreamConverter(ObjectMapper mapper) {
+        this.mapper = mapper;
+        this.medias = SUPPORTED_MEDIA_TYPES;
```
whiskeysierra: Why don't put the assignment to the field, if it's a constant?

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamConverter.java
```diff
@@ -0,0 +1,166 @@
...
+        // Conditional call because Jackson 2.7 does not support null contextClass anymore
+        // TypeVariable resolution will not work with Jackson 2.7, see SPR-13853 for more details
+        return (contextClass != null) ? tf.constructType(type, contextClass) : tf.constructType(type);
+    }
+
+    private boolean canRead(MediaType mediaType) {
+        if (mediaType == null) {
+            return true;
+        }
+        for (MediaType medias : getSupportedMediaTypes()) {
```
whiskeysierra:
```java
return getSupportedMediaTypes().anyMatch(mediaType::isCompatibleWith);
```

Should also do the trick.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamFilter.java
whiskeysierra: Ok, I get it now. FilterInputStream delegates skip to the delegate, instead of using the default implementaiton from InputStream. Unfortunate class design from the JDK.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamConverter.java
whiskeysierra: Or
```java
return mediaType == null  || getSupportedMediaTypes().anyMatch(mediaType::isCompatibleWith);
```

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamConverter.java
tkrop: You do not need to, but I will provide a solution that allows this.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamFilter.java
```diff
@@ -0,0 +1,95 @@
...
+ * ​⁣
+ */
+
+import java.io.FilterInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+class StreamFilter extends FilterInputStream {
+
+    private static final int JSON_SEQUENCE_SEPARATOR = 30;
```
whiskeysierra: Nitpicking, but you're not separating sequences but records inside a sequence.

riptide-stream/README.md
```diff
@@ -0,0 +1,26 @@
+# Riptide: Streams Extension
+
+
+*Riptide Stream* extension allows to capture arbitrary infinite object streams via Spring's [RestTemplate](https://spring.io/guides/gs/consuming-rest/). This includes infinite streaming format as application/x-json-stream and application/json-seq, but also streaming of simple finite lists/arrays of JSON objects. To enable streaming you only need to register the [stream converter](src/main/java/org/zalando/riptide/stream/StreamConverter.java) with Riptide and declare a route for your stream that is calling a the stream consumer as follows:
```
lukasniemeier-zalando: Why `RestTemplate`?

riptide-stream/README.md
```diff
@@ -0,0 +1,26 @@
...
+        rest.get("/repos/{org}/{repo}/contributors", "zalando", "riptide")
+                .accept(MediaType.APPLICATION_JSON)
+                .dispatch(series(),
+                       on(SUCCESSFUL).call(streamOf(User.class),
+                               forEach(user -> println(user.login + " (" + user.contributions + ")"))))
+                .get();
+    }
+}
+
+The unique entry point for all specific methods is the [Streams](src/main/java/org/zalando/riptide/stream/Streams.java) class. *Note:* The stream converter is an replacement to the default spring JSON converter that does not support streaming, and thus should be not registered together with it.
```
lukasniemeier-zalando: I do not understand the _Note_ part. The "default spring JSON converter" is the `MappingJackson..Converter` which is registered by Riptide by default? If registering both fails should this be prevented?

riptide-stream/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.zalando</groupId>
+        <artifactId>riptide-parent</artifactId>
+        <version>2.0.0-SNAPSHOT</version>
+    </parent>
+
+    <prerequisites>
+        <maven>3.0.4</maven>
```
lukasniemeier-zalando: Are we sure? :wink:

whiskeysierra:
+1

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamFilter.java
tkrop: As you like it ... see subsequent check in.

PR #130, opened 2016-07-14
Feature streams (code cleanup)

tkrop:


riptide-stream/README.md
```diff
@@ -1,7 +1,7 @@
 # Riptide: Streams Extension
 
 
-*Riptide Stream* extension allows to capture arbitrary infinite object streams via Spring's [RestTemplate](https://spring.io/guides/gs/consuming-rest/). This includes infinite streaming format as application/x-json-stream and application/json-seq, but also streaming of simple finite lists/arrays of JSON objects. To enable streaming you only need to register the [stream converter](src/main/java/org/zalando/riptide/stream/StreamConverter.java) with Riptide and declare a route for your stream that is calling a the stream consumer as follows:
+*Riptide Stream* extension allows to capture arbitrary infinite object streams via Spring's [Rest Template](https://spring.io/guides/gs/consuming-rest/). This includes infinite streaming format as application/x-json-stream and application/json-seq, but also streaming of simple finite lists/arrays of JSON objects. To enable streaming you only need to register the [stream converter](src/main/java/org/zalando/riptide/stream/StreamConverter.java) with Riptide and declare a route for your stream that is calling a the stream consumer as follows:
```
lukasniemeier-zalando: Why `RestTemplate`?

riptide-stream/src/test/java/org/zalando/riptide/stream/EnforceCoverageTest.java
```diff
@@ -37,14 +37,14 @@
 
     @Test
     public void shouldUseStreamsConstructor() {
-        new Streams();
+        assertNotNull(new Streams());
```
lukasniemeier-zalando: Hm :wink: see test below

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamSpliterator.java
```diff
@@ -47,6 +47,7 @@ public boolean tryAdvance(Consumer<? super T> action) {
         try {
             JsonToken token = parser.nextToken();
             if (token == null) {
+                parser.close();
```
whiskeysierra: We need to close that also if somebody closes the stream himself. Do we have a test for that?

riptide-stream/src/main/java/org/zalando/riptide/stream/Streams.java
```diff
@@ -56,23 +56,50 @@
  */
 public final class Streams {
 
+    /** Default stream character encoding. */
     public static final Charset DEFAULT_CHARSET = Charset.forName("UTF-8");
```
whiskeysierra: Is it necessary to have this as a public API? Also, you can just use StandardCharsets.UTF_8

riptide-stream/src/main/java/org/zalando/riptide/stream/Streams.java
```diff
@@ -81,16 +108,27 @@
...
             return (Exception) super.getCause();
         }
     }
 
+    /**
+     * Creates {@link ThrowingConsumer stream consumer} for given {@link ThrowingConsumer element consumer}. Used to
+     * standardly wrap a single entity consumer function in a stream consumer function as follows:
+     * 
+     * <pre>
+     *     on(...).call(streamOf(...), forEach((element) -> { println(element); }))
```
whiskeysierra: Parenthesis are not needed around *element*.

riptide-stream/src/test/java/org/zalando/riptide/stream/EnforceCoverageTest.java
```diff
@@ -37,14 +37,14 @@
...
     }
 
     @Test
     public void shouldUseStreamConverterConstructor() {
-        Streams.streamConverter();
-        Streams.streamConverter(null);
-        Streams.streamConverter(null, null);
+        streamConverter();
```
whiskeysierra: I still strongly believe that those should be covered by real tests. We only cover useless constructors in here, i.e. hidden constructors from static factory classes.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamSpliterator.java
tkrop: I do not understand. We only close automatically when reaching the end. When we never reach the end, I do not now how I should close. The Spliterator itself is not closeable and will not be close by the stream support. I agree, we are something missing here and that's why I added this line here.

riptide-stream/src/main/java/org/zalando/riptide/stream/Streams.java
tkrop: No. I will update that. Was not aware of this support.

riptide-stream/src/test/java/org/zalando/riptide/stream/EnforceCoverageTest.java
tkrop: Okay,  I only did this to make my IDE happy ;).

riptide-stream/src/test/java/org/zalando/riptide/stream/EnforceCoverageTest.java
tkrop: On a black box, yes! But this is a white box and I know testing this would be highly redundant or just checking for assertNotNull as above.

whiskeysierra: No, your test should verify that the constructor is doing what it's supposed to do. E.g. that it allows to override media types (constructor call + assert on getter).

whiskeysierra: > testing this would be highly redundant

If that were true than coverage wouldn't complain.

riptide-stream/src/test/java/org/zalando/riptide/stream/EnforceCoverageTest.java
whiskeysierra: Your IDE is weird! :boom:

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamSpliterator.java
whiskeysierra: Ok, so a Stream itself is Closeable, which means there has to be a way to hook into it. I read a little bit about it and it looks like terminal operations (forEach, collect, ...) actually do **not** close the stream. So in general clients should always close them.

whiskeysierra: This should do the trick:

```java
    /**
     * Returns an equivalent stream with an additional close handler.  Close
     * handlers are run when the {@link #close()} method
     * is called on the stream, and are executed in the order they were
     * added.  All close handlers are run, even if earlier close handlers throw
     * exceptions.  If any close handler throws an exception, the first
     * exception thrown will be relayed to the caller of {@code close()}, with
     * any remaining exceptions added to that exception as suppressed exceptions
     * (unless one of the remaining exceptions is the same exception as the
     * first exception, since an exception cannot suppress itself.)  May
     * return itself.
     *
     * <p>This is an <a href="package-summary.html#StreamOps">intermediate
     * operation</a>.
     *
     * @param closeHandler A task to execute when the stream is closed
     * @return a stream with a handler that is run if the stream is closed
     */
    S onClose(Runnable closeHandler);
```

whiskeysierra: https://docs.oracle.com/javase/8/docs/api/java/util/stream/BaseStream.html#onClose-java.lang.Runnable-

whiskeysierra: I'd also like to have the close operation on the same logical layer as the creation of the parser (the same applies to io streams), i.e. I'd do the close in the method that creates the parser.

tkrop: Agreed!

riptide-stream/src/test/java/org/zalando/riptide/stream/EnforceCoverageTest.java
tkrop: My test do. And they implicitly check the defaults, too. Only, the above public support functions are tested separately. The best I would offer for them is checking that call are not throwing unexpected exceptions and that the result is an actual object.

whiskeysierra: Why not use them in your tests directly then, instead of calling the constructor?

tkrop: Because I have the weird opinion that a unit test should use the component or testing instead of some hiding support class.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamSpliterator.java
tkrop: Testing close is pain in the ass.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamConverter.java
```diff
@@ -45,11 +45,10 @@
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.type.TypeFactory;
 
-class StreamConverter<T> implements GenericHttpMessageConverter<T> {
+final class StreamConverter<T> implements GenericHttpMessageConverter<T> {
 
     private static final List<MediaType> DEFAULT_MEDIA_TYPES =
-            Arrays.asList(APPLICATION_JSON_SEQ,
-                    APPLICATION_X_JSON_STREAM);
+            Arrays.asList(APPLICATION_JSON_SEQ, APPLICATION_X_JSON_STREAM);
```
whiskeysierra: This returns a fixed-size, **mutable** list which you're then returning in `getSupportedMediaTypes`, so someone could change it.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamConverter.java
```diff
@@ -106,49 +105,52 @@ public boolean canWrite(Type type, Class<?> clazz, MediaType mediaType) {
...
         }
     }
 
-    private StreamSpliterator<Object> split(JavaType javaType, HttpInputMessage inputMessage)
+    @SuppressWarnings("unchecked")
+    private T stream(final JavaType javaType, final InputStream stream)
             throws IOException, JsonParseException {
+        final JsonParser parser = mapper.getFactory().createParser(stream);
+        final StreamSpliterator<Object> split = new StreamSpliterator<>(javaType, parser, mapper);
+        return (T) StreamSupport.stream(split, false).onClose(split::close);
```
whiskeysierra: Why adding a close method to the Spliterator? We're creating the parser here, so I'd expect:

```java
return (T) StreamSupport.stream(split, false).onClose(parser::close);
```

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamSpliterator.java
```diff
@@ -86,4 +86,13 @@ public long estimateSize() {
     public int characteristics() {
         return ORDERED | IMMUTABLE;
     }
+
+    @Override
+    public void close() throws UncheckedIOException {
```
jhorstmann: `AutoCloseable` allows to throw any exception so there should be no need to wrap a checked exception here. But as Willi said, you could just use the `JsonParser#close` method as the `onClose` handler.

riptide-stream/README.md
```diff
@@ -1,7 +1,7 @@
 # Riptide: Streams Extension
```
jhorstmann: Could document that this follows a slightly more liberal interpretation of the `json-seq` spec, as it will ignore record separators anywhere in the content instead of requiring them between separate objects.

The `x-json-stream` type does not have a definite specification, but I would understand it as json values separated by optional newlines.

jhorstmann: Could document that this follows a slightly more liberal interpretation of the `json-seq` spec, as it will ignore record separators anywhere in the content instead of requiring them between separate objects.

The `x-json-stream` type does not have a definite specification, but I would understand it as json values separated by optional newlines.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamSpliterator.java
whiskeysierra: I'm actually not 100% sure that calling JsonParser#close is enough. Depending on configuration it may not close the stream, so we may need to do it here.

tkrop: JsonParser will close the stream correctly per default, we may go with this solution. Wrapping it in the Spliterator was the idea to handle the IOException problem, that is not allowed to appear on the stream.

riptide-stream/README.md
tkrop: Yes you are right, but I didn't like users to depend on this. It's result of the problem, that I could not find an easy solution. removing the value using the JsonParser itself, but I might find in the future.

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamConverter.java
tkrop: Good point, but I have the problem, that the parser emits a checked exception, the stream is not allowed to catch. I need to wrap it, either here or in the spliterator. I assume here is cleaner, so I will change it.

tkrop: But now it is harder to test ...

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamSpliterator.java
whiskeysierra: > JsonParser will close the stream correctly per default

Yes, but if somebody changes the configuration we are not behaving correctly and since we are not exposing a way to close the stream the client is unable to workaround that issue.

tkrop: I do not think this is a problem at all.

The input stream will be closed whenever the connection is closed. This is part of the lower layer protocols. As long as we take care that we close the stream/connection, when some error happens in our receiver, we will have no stale resources hanging.

As far as I see, this is taken care of by the current implementation.

PR #137, opened 2016-08-03
Added support for query parameters

whiskeysierra:


riptide-core/src/main/java/org/zalando/riptide/Requester.java
```diff
@@ -30,8 +32,14 @@
 
 public abstract class Requester extends Dispatcher {
 
+    private final Multimap<String, String> query = LinkedHashMultimap.create();
     private final HttpHeaders headers = new HttpHeaders();
 
+    public final Requester query(final String name, final String value) {
```
AlexanderYastrebov: `param`?

AlexanderYastrebov:
I thought UrlTemplate does this substitution

riptide-core/src/main/java/org/zalando/riptide/Requester.java
lukasniemeier-zalando: `queryParam`?

whiskeysierra: Yes, if we change it then to `queryParam`. `param` itself has too many possible meanings in the context of HTTP: query param, post param, form param, ...

whiskeysierra:
> I thought UrlTemplate does this substitution

No, UriTemplate is only doing parameter replacement, but it doesn't have support for query parameters as far as I can tell.

tkrop:
@whiskeysierra you should try it. DefaultUriTemplateHandler does substitute query parameter too.
Try it out:
```
DefaultUriTemplateHandler h = new DefaultUriTemplateHandler();
        h.setBaseUrl("http://xample.com");
        URI uri = h.expand("/path/{a}/{b}/x?y={c}&z={d}", "r", "s", "t", "u");
        System.out.println(uri);
        uri = h.expand("/path/{a}/{b}/x?y={c}&z={d}",
                (Map<String,?>) ImmutableMap.<String, Object>builder().put("b", "r").put("a", "s").put("d", "t").put("c", "u").build());
        System.out.println(uri);
```

So this feature is mostly redundant to the uri-template support. In my opinion, the only use-case with additional value would require the interface `queryParams(final Multimap<String, String> query)`, that allows to ad any number of query parameters defined by the client.

whiskeysierra:
Yes, that works but It doesn't feel very natural to me, and the overload makes sense.

lukasniemeier-zalando:
*As* a riptide user
*I do not want* to fiddle with `UriTemplateHandler` in most cases
*So that* I build my URL in one place

lukasniemeier-zalando:
:+1:

riptide-core/src/main/java/org/zalando/riptide/Rest.java
```diff
@@ -120,30 +124,36 @@ public final Requester trace(final URI url) {
 
     private Requester execute(final HttpMethod method, final String uriTemplate,
             final Object... uriVariables) {
-        return execute(method, uriTemplateHandler.expand(uriTemplate, uriVariables));
+        return new ListenableFutureRequester(method, fromUriString(uriTemplate), uriVariables);
```
tkrop: The naming is bad. Actually, I have the feeling the methods of this class should be integrated into the Requester itself.

lukasniemeier-zalando:
:+1:

PR #138, opened 2016-08-03
Fixed consumption of response body if closing it

whiskeysierra:


riptide-httpclient/src/main/java/org/zalando/riptide/httpclient/RestAsyncClientHttpResponse.java
```diff
@@ -0,0 +1,79 @@
...
+        return response.getBody();
+    }
+
+    @Override
+    public HttpHeaders getHeaders() {
+        return response.getHeaders();
+    }
+
+    @Override
+    @SneakyThrows
```
jhorstmann: Just because of your code coverage requirements, right? To work around that you could have a helper method taking a `Callable` and have that wrap or propagate exceptions.

whiskeysierra: > Just because of your code coverage requirements, right?

Yes :stuck_out_tongue_winking_eye: 

Actually, now that I'm thinking about it, our riptide-stream extension is operating on the stream directly and trying to close it, so this won't even fix the real problem if `ChunkedInputStream#close` is the root cause.

riptide-httpclient/src/main/java/org/zalando/riptide/httpclient/RestAsyncClientHttpResponse.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Override
+    public InputStream getBody() throws IOException {
+        return new FilterInputStream(response.getBody()) {
+            @Override
+            public void close() throws IOException {
+                // intentionally left blank
+                // Some implementations (looking at you, org.apache.http.impl.io.ChunkedInputStream) actually try
+                // to completely consume the content upon close which of course doesn't work for endless streams.
+                // This will cause the connection to not be released back to the connection pool.
```
lukasniemeier-zalando: "This" in this case means your empty implementation?

tkrop: I think this implementation is wrong. I have debugged this situation and found out, that the connection is still pending until it times out. To finish without reading, the correct action is to implement it as follows:

```
if (in instanceof EofSensorInputStream) {
      ((EofSensorInputStream) in).abortConnection();
} else {
       in.close();
}
```

whiskeysierra:
Will be part of changes coming from @tkrop

PR #140, opened 2016-08-15
Feature: nakadi stream support

tkrop:
The test case for client exceptions in streams needs special REST client driver support that must be enabled by building from the branch https://github.com/tkrop/rest-driver/tree/feature/stream-and-lamda-support.


riptide-core/src/main/java/org/zalando/riptide/RestBuilder.java
```diff
@@ -84,6 +90,14 @@ public static RestConfigurer simpleRequestFactory(final ExecutorService executor
         };
     }
 
+    @Override
+    public RestBuilder clone() {
```
whiskeysierra: Why is this needed? Shouldn't the builder be re-usable by default?

riptide-core/src/main/java/org/zalando/riptide/RestBuilder.java
```diff
@@ -84,6 +90,14 @@ public static RestConfigurer simpleRequestFactory(final ExecutorService executor
...
+        return new RestBuilder()
+                .requestFactory(requestFactory)
+                .converters(converters)
+                .baseUrl(baseUrl);
+    }
+
     public Rest build() {
         return new Rest(requestFactory, converters(), baseUrl);
```
whiskeysierra: If we make a copy of the list here the builder can be re-used without cloning it.

riptide-core/src/main/java/org/zalando/riptide/Route.java
```diff
@@ -22,28 +31,30 @@
...
+        };
     }
 
     static Route call(final ThrowingConsumer<ClientHttpResponse> consumer) {
-        return (response, reader) ->
+        return (response, reader) -> {
+            try {
                 consumer.accept(response);
+            } finally {
+                response.close();
```
whiskeysierra: If we somehow wrap this in a try-with-resource block we would get the additional support of surpressed exceptions, right?

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/Batch.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+import java.util.List;
+
+import lombok.Value;
+
+@Value
+final class Batch {
+
+    Cursor cursor;
```
whiskeysierra: Cursor should be of type JsonNode because we don't care about the actual content. Maybe a package private `@JsonValue JsonNode getValue()` to hide that from clients.

riptide-httpclient/src/main/java/org/zalando/riptide/httpclient/RestAsyncClientHttpResponse.java
```diff
@@ -0,0 +1,81 @@
...
+        };
+    }
+
+    @Override
+    public HttpHeaders getHeaders() {
+        return response.getHeaders();
+    }
+
+    @Override
+    @SneakyThrows
```
jhorstmann: Sneaking around the code coverage requirements again :)

riptide-nakadi/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+        </dependency>
+
+        <!-- rest -->
+        <dependency>
+            <groupId>com.github.rest-driver</groupId>
+            <artifactId>rest-client-driver</artifactId>
+        </dependency>
+    </dependencies>
+ 
+    <distributionManagement>
```
jhorstmann: This could be configured in the parent

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
```diff
@@ -0,0 +1,179 @@
...
+    private static final String SESSION_HEADER = "X-Nakadi-SessionId"; // TODO should this be X-Nakadi-StreamId?
+
+    private static final Binding<Series> ON_FAILURE_BINDING =
+            anySeries().call(dispatch(contentType(),
+                    on(PROBLEM).call(ThrowableProblem.class, propagate()),
+                    Bindings.anyContentType().call(invalidContent())));
+
+    private static ThrowingConsumer<ClientHttpResponse> invalidContent() {
+        return response -> {
+            InputStreamReader input = new InputStreamReader(response.getBody());
```
jhorstmann: Should specify charset.

Instead of reading line by line and then joining you could also use guava `CharStream.toString(Readable)`.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/Event.java
```diff
@@ -0,0 +1,72 @@
...
+
+    }
+
+    Metadata metadata;
+
+    String dataType;
+
+    DataOperation dataOp;
+
+    ObjectNode data;
```
jhorstmann: Do I see this correctly that you only support datachange events, and that users would need to databind from the `ObjectNode` to their own data model themselves?

riptide-core/src/main/java/org/zalando/riptide/RestBuilder.java
tkrop: Only used at one location. Will remove it until more use cases are visible.

riptide-core/src/main/java/org/zalando/riptide/Route.java
tkrop: This would be in the anonymous dispatcher in the Requester.
```
future.addCallback(response -> {
                    try { ...
```
Doing it here would close the stream in any case. I think this would be okay for all use cases. Anyhow, I can't see how this would result in suppressed exceptions without having code coverage problems.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/Batch.java
tkrop: No, we care. We need to commit this cursor at the end.

whiskeysierra: Yes, we need it, but we don't care about the actual content/properties.

riptide-httpclient/src/main/java/org/zalando/riptide/httpclient/RestAsyncClientHttpResponse.java
tkrop: No, left over artifact of code change. Thanks for detecting it.

riptide-core/src/main/java/org/zalando/riptide/Route.java
whiskeysierra: Using try-with-resource will add any exception during close to the list of supressed exceptions (`Throwable.getSuppressed`) for further analysis (will be included in logfiles, iirc).

riptide-nakadi/pom.xml
tkrop: Yes, it is already there. I will remove this.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
tkrop: Although an option. Better is @whiskeysierra suggested:
```
new String(ByteStreams.toByteArray(response.getBody()), charset)
```
I like the above option, because it is without guava :)

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/Event.java
tkrop: No, the support for datachange events should already be optional. The fields are available but must not be filled in by the stream.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
jhorstmann: Isn't `ByteStreams` also from guava?

riptide-core/src/main/java/org/zalando/riptide/Requester.java
```diff
@@ -1,5 +1,7 @@
 package org.zalando.riptide;
 
+import static org.zalando.riptide.TryWith.tryWith;
```
lukasniemeier-zalando: ?

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/Event.java
whiskeysierra: I'd prefer to have `T` here (and the class to be generic) so clients don't have to do the mapping themselves.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/Event.java
```diff
@@ -0,0 +1,72 @@
...
+
+import com.fasterxml.jackson.annotation.JsonValue;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import lombok.Value;
+
+@Value
+public class Event {
+
+    public enum DataOperation {
```
whiskeysierra: I'd make those inner classes top-level ones.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/Event.java
```diff
@@ -0,0 +1,72 @@
...
+import java.time.OffsetDateTime;
+import java.util.UUID;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import lombok.Value;
+
+@Value
+public class Event {
```
whiskeysierra: DataChangeEvent?

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/Event.java
```diff
@@ -0,0 +1,72 @@
...
+
+        String flowId;
+
+    }
+
+    Metadata metadata;
+
+    String dataType;
+
+    DataOperation dataOp;
```
whiskeysierra:
```java
@JsonProperty("data_op")
Operation operation;
```

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/Event.java
whiskeysierra: Just Operation?!

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/Event.java
whiskeysierra: > No, the support for datachange events should already be optional. The fields are available but need not to be filled in by the stream.

I wouldn't mix those different types of events in a single class.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
```diff
@@ -0,0 +1,183 @@
...
+            Charset charset = Optional.ofNullable(response.getHeaders().getContentType())
+                    .map(MediaType::getCharSet)
+                    .orElse((StandardCharsets.ISO_8859_1));
+            String body = new String(ByteStreams.toByteArray(response.getBody()), charset);
+            throw new IOException("Unsupported content: " + response.getRawStatusCode() +
+                    " - " + response.getStatusText() + "\n" + response.getHeaders() + "\n" + body);
+        };
+    }
+
+    private static ThrowingConsumer<ThrowableProblem> propagate() {
```
whiskeysierra: Not needed, the propagate from the core will do the same thing.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
```diff
@@ -0,0 +1,183 @@
...
+        return entity -> {
+            throw entity;
+        };
+    }
+
+    static class GatewayException extends RuntimeException {
```
whiskeysierra: What's the use case for this? How does it help if it's not public?

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
```diff
@@ -0,0 +1,183 @@
...
+import org.zalando.riptide.Rest;
+import org.zalando.riptide.ThrowingConsumer;
+import org.zalando.riptide.capture.Capture;
+
+import com.google.common.io.ByteStreams;
+
+@Component
+public class NakadiGateway {
+
+    public static final MediaType PROBLEM = parseMediaType("application/problem+json");
```
whiskeysierra: We really should have a constant for that media type in the problem library.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
```diff
@@ -0,0 +1,183 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(NakadiGateway.class);
+    private static final String SESSION_HEADER = "X-Nakadi-SessionId"; // TODO should this be X-Nakadi-StreamId?
+
+    private static final Binding<Series> ON_FAILURE_BINDING =
```
whiskeysierra: I'd use a Route here.

whiskeysierra: Doesn't have to be static, since the class is meant to be a singleton.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
```diff
@@ -0,0 +1,183 @@
...
+import org.zalando.problem.ThrowableProblem;
+import org.zalando.riptide.Binding;
+import org.zalando.riptide.Bindings;
+import org.zalando.riptide.Rest;
+import org.zalando.riptide.ThrowingConsumer;
+import org.zalando.riptide.capture.Capture;
+
+import com.google.common.io.ByteStreams;
+
+@Component
```
whiskeysierra: I'd not use those Component/Autowired in here. Clients should wire them manually, since we can't have meaningful defaults.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    private final Rest rest;
+
+    @Autowired
+    public NakadiGateway(final Rest rest) {
+        this.rest = rest;
+    }
+
+    public void stream(final String owningApplication, final List<String> eventTypes, final EventConsumer consumer)
```
whiskeysierra: Owning application should be a field/constructor argument, imho.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
```diff
@@ -0,0 +1,183 @@
...
+            throws InterruptedException, ExecutionException, IOException {
+        stream(subscribe(new Subscription(null, owningApplication, eventTypes)), consumer);
+    }
+
+    private void stream(final Subscription subscription, final EventConsumer consumer)
```
whiskeysierra: Since we're subscribing first and streaming later this method should be after `subscribe`.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
whiskeysierra: I'd love to have this method generic, including the EventConsumer interface.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
```diff
@@ -0,0 +1,183 @@
...
+
+    private ThrowingConsumer<ResponseEntity<Stream<Batch>>> process(final Subscription subscription,
+            final EventConsumer consumer) {
+        return entity -> {
+            @SuppressWarnings("resource")
+            final Stream<Batch> stream = entity.getBody();
+            try {
+                final HttpHeaders headers = entity.getHeaders();
+                stream.map(batch -> consume(consumer, batch))
+                        .filter(cursor -> cursor != null)
```
whiskeysierra:
```java
.filter(Objects::nonNull)
```

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
```diff
@@ -0,0 +1,183 @@
...
+    private Cursor consume(final EventConsumer consumer, final Batch batch) {
+        Cursor cursor = batch.getCursor();
+        if (cursor == null) {
+            LOG.debug("invalid batch: [{}]", batch);
+            return cursor;
+        }
+        List<Event> events = batch.getEvents();
+        if (events == null || events.isEmpty()) {
+            LOG.debug("empty batch: [{}]", batch);
+            return cursor;
```
whiskeysierra: I think returning null would be cleaner here, since there is nothing to commit, is there? @AlexanderYastrebov Thoughts, since you talked to the Nakadi guys about this?

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
```diff
@@ -0,0 +1,183 @@
...
+        }
+        events.forEach(event -> consumer.consume(event));
```
whiskeysierra:
```java
events.forEach(consumer::consume);
```

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
```diff
@@ -0,0 +1,183 @@
...
+        return cursor;
+    }
+
+    private Cursor[] commit(final Subscription subscription, final HttpHeaders headers, final Cursor... cursor) {
```
whiskeysierra: Why not a `List<Cursor>`?

whiskeysierra: Why even having a return value at all? You're not using it anywhere, are you?

whiskeysierra: Instead of `HttpHeaders` I feel like a `String streamId` would be more honest in terms of what the method actually needs.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
```diff
@@ -0,0 +1,183 @@
...
+            rest.put("/subscriptions/{id}/cursors", subscription.getId())
+                    .contentType(MediaType.APPLICATION_JSON)
+                    .header(SESSION_HEADER, headers.getFirst(SESSION_HEADER))
+                    .body(cursor)
+                    .dispatch(series(),
+                            on(SUCCESSFUL).call(Cursor[].class, capture),
+                            ON_FAILURE_BINDING)
+                    .get();
+        } catch (ExecutionException ex) {
+            throw new GatewayException(ex.getCause());
```
whiskeysierra: Why wrapping everything here?

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
```diff
@@ -0,0 +1,183 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(NakadiGateway.class);
+    private static final String SESSION_HEADER = "X-Nakadi-SessionId"; // TODO should this be X-Nakadi-StreamId?
+
+    private static final Binding<Series> ON_FAILURE_BINDING =
+            anySeries().call(dispatch(contentType(),
+                    on(PROBLEM).call(ThrowableProblem.class, propagate()),
+                    Bindings.anyContentType().call(invalidContent())));
+
+    private static ThrowingConsumer<ClientHttpResponse> invalidContent() {
```
whiskeysierra: `onUnknownErrorFormat`?

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
tkrop: No, it will wrap it in an IOException, which I didn't like here to have.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
tkrop: Okay, it is not feature complete at the moment, and in a first approach I was not sure, whether I should unwrap this exception before giving the resulting exception to the consumer.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
tkrop: There is currently nothing that hinders clients to build multiple NakadiGateways, and I see no reason to do so.

> I'd use a Route here.

It's still a binding, so I'm preferring it to name it that way.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
tkrop: Got me. Too much copy&past.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
tkrop: > Owning application should be a field/constructor argument, imho.

Yes semantically agreed. And we have to include client_group_id here.

> I'd love to have this method generic, including the EventConsumer interface.

When we substitute the generic event against specific event types, this makes sense. As long as we do not plane to translate an generic event into a specific type, I would prefer to leave it that way.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
tkrop: I prefer to have methods that continue each other to keep close together, especially if they have the same name. How about putting subscribe above streams, because you have to subscribe before streaming at all.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
tkrop: True, didn't pushed this change yesterday.

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
tkrop: > Why even having a return value at all? You're not using it anywhere, are you?

That's why I have not had a plan for this now. I even do not understand the reason why we have a list of coursers, while a batch contains only a single one...

riptide-nakadi/src/main/java/org/zalando/riptide/nakadi/NakadiGateway.java
tkrop: Yes something in this direction would be good.

riptide-core/src/main/java/org/zalando/riptide/NoRouteException.java
```diff
@@ -36,8 +43,26 @@ public NoRouteException(final ClientHttpResponse response) throws IOException {
...
-        return String.format("Unable to dispatch response (%d %s, Content-Type: %s)",
-                response.getRawStatusCode(), response.getStatusText(), response.getHeaders().getContentType());
+        return String.format("Unable to dispatch response: %d - %s\n%s\n%s",
+                response.getRawStatusCode(), response.getStatusText(), response.getHeaders(),
+                readStartOfBody(response, MAX_BODY_BYTES_TO_READ));
+    }
+
+    private static String readStartOfBody(final ClientHttpResponse response, final int length) throws IOException {
+        final InputStream stream = response.getBody();
+        if (stream != null) {
```
jhorstmann: stream should be closed in a finally block

AlexanderYastrebov:
Why is this made as a Riptide module? It is Nakadi-specific and should be separate library.

PR #148, opened 2016-08-18
Core improvements and bug-fixes for nakadi stream support

tkrop:


riptide-core/src/main/java/org/zalando/riptide/NoRouteException.java
```diff
@@ -36,8 +43,26 @@ public NoRouteException(final ClientHttpResponse response) throws IOException {
     }
 
     private static String formatMessage(final ClientHttpResponse response) throws IOException {
-        return String.format("Unable to dispatch response (%d %s, Content-Type: %s)",
-                response.getRawStatusCode(), response.getStatusText(), response.getHeaders().getContentType());
+        return String.format("Unable to dispatch response: %d - %s\n%s\n%s",
+                response.getRawStatusCode(), response.getStatusText(), response.getHeaders(),
+                readStartOfBody(response, MAX_BODY_BYTES_TO_READ));
```
lukasniemeier-zalando: Are you doing this so that the response is consumed? :wink:

riptide-core/src/main/java/org/zalando/riptide/RestBuilder.java
```diff
@@ -56,6 +56,11 @@ public RestBuilder defaultConverters() {
         return converters(Converters.DEFAULT);
     }
 
+    public RestBuilder clearConverters() {
```
lukasniemeier-zalando: Why would you need that?

riptide-core/src/test/java/org/zalando/riptide/TryWithTest.java
```diff
@@ -0,0 +1,187 @@
...
+import java.io.Closeable;
+import java.nio.charset.CharacterCodingException;
+
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.mockito.Mockito;
+
+public class TryWithTest {
```
lukasniemeier-zalando: Static imports for the `Mockito` stuff?

riptide-core/src/main/java/org/zalando/riptide/NoRouteException.java
whiskeysierra: No, this is for debugging/logging.

riptide-core/src/test/java/org/zalando/riptide/HandlesIOExceptionTest.java
```diff
@@ -48,7 +48,7 @@ public void shouldHandleExceptionDuringRequestCreation() throws URISyntaxExcepti
             throw new IOException("Could not create request");
         };
 
-        Rest.builder().requestFactory(factory).build()
+        Rest.builder().requestFactory(factory).defaultConverters().build()
```
whiskeysierra: Are you doing this to use the method at least once?

riptide-core/src/test/java/org/zalando/riptide/TryWithTest.java
```diff
@@ -0,0 +1,187 @@
...
+
+    @Test
+    public void shouldCloseAfterRunningWithException() throws Exception {
+        exception.expect(IllegalStateException.class);
+
+        Closeable closable = mock(Closeable.class);
+        ThrowingRunnable consumer = mock(ThrowingRunnable.class);
+
+        Mockito.doThrow(new IllegalStateException()).when(consumer).run();
+        Mockito.doNothing().when(closable).close();
```
whiskeysierra: That is the default, no need to specify that.

riptide-core/src/test/java/org/zalando/riptide/TryWithTest.java
```diff
@@ -0,0 +1,187 @@
...
+
+    @Rule
+    public final ExpectedException exception = ExpectedException.none();
+
+    @Test
+    public void shouldCloseAfterRunningWithoutException() throws Exception {
+        Closeable closable = mock(Closeable.class);
+        ThrowingRunnable consumer = mock(ThrowingRunnable.class);
+
+        Mockito.doNothing().when(consumer).run();
```
whiskeysierra: Not needed.

riptide-core/src/test/java/org/zalando/riptide/TryWithTest.java
```diff
@@ -0,0 +1,187 @@
...
+        Mockito.doNothing().when(closable).close();
```
whiskeysierra: Same here.

riptide-core/src/test/java/org/zalando/riptide/TryWithTest.java
```diff
@@ -0,0 +1,187 @@
...
+
+        tryWith(closable, consumer);
+
+        Mockito.verify(consumer, times(1)).run();
```
whiskeysierra: `times(1)` is also the default.

riptide-core/src/test/java/org/zalando/riptide/TryWithTest.java
whiskeysierra: A general rule of thumb: stubbing (`when`) and verification (`verify`) should be seen as mutually exclusive. 99% of tests should only ever use one xor the other. Very, very rarely you need to use both for the same mock.

riptide-core/src/test/java/org/zalando/riptide/TryWithTest.java
```diff
@@ -0,0 +1,187 @@
...
+        Closeable closable = mock(Closeable.class);
+        ThrowingRunnable consumer = mock(ThrowingRunnable.class);
+
+        Mockito.doThrow(new IllegalStateException()).when(consumer).run();
+        Mockito.doNothing().when(closable).close();
+
+        try {
+            tryWith(closable, consumer);
+        } finally {
+            Mockito.verify(consumer, times(1)).run();
```
whiskeysierra: Not needed for the test.

riptide-core/src/test/java/org/zalando/riptide/TryWithTest.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    @Test
+    public void shouldCloseAfterRunningAndExposeExceptionWhenFailintToClose() throws Exception {
+        exception.expect(CharacterCodingException.class);
+
+        Closeable closable = mock(Closeable.class);
+        ThrowingRunnable consumer = mock(ThrowingRunnable.class);
+
+        Mockito.doNothing().when(consumer).run();
```
whiskeysierra: Not needed.

riptide-core/src/test/java/org/zalando/riptide/TryWithTest.java
```diff
@@ -0,0 +1,187 @@
...
+        Mockito.doThrow(new CharacterCodingException()).when(closable).close();
+
+        try {
+            tryWith(closable, consumer);
+        } finally {
+            Mockito.verify(consumer, times(1)).run();
+            Mockito.verify(closable, times(1)).close();
```
whiskeysierra: Not needed.

riptide-core/src/test/java/org/zalando/riptide/TryWithTest.java
```diff
@@ -0,0 +1,187 @@
...
+        Mockito.doThrow(new CharacterCodingException()).when(closable).close();
+
+        try {
+            tryWith(closable, consumer);
+        } catch (Exception ex) {
+            Assert.assertThat(ex.getSuppressed().length, is(equalTo(1)));
+            Assert.assertThat(ex.getSuppressed()[0], is(instanceOf(CharacterCodingException.class)));
+            throw ex;
+        } finally {
+            Mockito.verify(consumer, times(1)).run();
```
whiskeysierra: Both not needed.

riptide-core/src/test/java/org/zalando/riptide/TryWithTest.java
```diff
@@ -0,0 +1,187 @@
...
+            Mockito.verify(closable, times(1)).close();
+        }
+    }
+
+    @Test
+    public void shouldCloseAfterAcceptingWithoutException() throws Exception {
+        Closeable closable = mock(Closeable.class);
+        ThrowingConsumer<?> consumer = mock(ThrowingConsumer.class);
+
+        Mockito.doNothing().when(consumer).accept(null);
```
whiskeysierra: That's the default.

riptide-httpclient/src/test/java/org/zalando/riptide/httpclient/RestAsyncClientHttpResponseBodyTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+        try (RestAsyncClientHttpResponse unit = new RestAsyncClientHttpResponse(response)) {
+            unit.getBody().close();
+        }
+
+        verify(stream, times(1)).close();
+    }
+
+    @Test
+    public void shouldCallCloseAndAbortOnConnectionReleaseTrigger() throws IOException {
```
whiskeysierra: The name of the tests suggests that there should also be a verification of the close method.

riptide-core/src/main/java/org/zalando/riptide/RestBuilder.java
tkrop: Yup, left over artifact from intermediate state.

riptide-core/src/test/java/org/zalando/riptide/HandlesIOExceptionTest.java
tkrop: You got it! Not my idea - was @whiskeysierra - moved it only to this location.

riptide-httpclient/src/test/java/org/zalando/riptide/httpclient/RestAsyncClientHttpResponseBodyTest.java
tkrop: Oh yes, it should. I will add this.

tkrop:
So, hopefully now it is  cleaned up and ready for merge.

tkrop:
Finally, also all `times(1)` removed.

riptide-core/src/test/java/org/zalando/riptide/tryit/TryEnsureTest.java
```diff
@@ -0,0 +1,128 @@
...
+import java.io.IOException;
+import java.nio.charset.CharacterCodingException;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.zalando.riptide.ThrowingRunnable;
+import org.zalando.riptide.ThrowingSupplier;
+
+public class TryEnsureTest {
```
whiskeysierra: This test and the other one still use stubbing and verification in an unnecessarily verbose fashion.

riptide-core/src/test/java/org/zalando/riptide/tryit/TryEnsureTest.java
```diff
@@ -0,0 +1,128 @@
...
+        this.exception.expect(IOException.class);
+        this.exception.expectCause(instanceOf(Exception.class));
+
+        ThrowingRunnable runnable = mock(ThrowingRunnable.class);
+
+        doThrow(new Exception()).when(runnable).run();
+        try {
+            ensureIOException(runnable);
+        } finally {
+            verify(runnable).run();
```
whiskeysierra: No need to verify stuff like this. You're already expecting an exception. If your runnable wouldn't be called there would be no exception. There is no way that this verify is catching any bug that you wouldn't already see with the expected exception part.

riptide-core/src/main/java/org/zalando/riptide/tryit/TryWith.java
```diff
@@ -0,0 +1,91 @@
+package org.zalando.riptide.tryit;
```
whiskeysierra: Why the package?

tkrop: Do you think it should stay in the main directory? I thought it might clutter the public visible API. In general I have the feeling, that the concept of Throwing* and Try* is more generic concept that riptide should import it from another project.

riptide-core/src/test/java/org/zalando/riptide/tryit/TryEnsureTest.java
tkrop: You are only right for white box testing, while I prefer here contract testing. While this might imply redundancy on the current implementation, it is more verbose and reliable against changes.

whiskeysierra: This has nothing to do with white-box vs. black-box. The contract of your method says that if it throws so should ensureIOException. That is what your would actually assert if you drop the verify. If you leave the verify in there you are expecting an exception and that your method was called. It's redundant and it's not more reliable against changes. It's just a fragile and harder to maintain test because you're doing too much. It's another good rule of thumb to only ever assert one *thing* per test.

tkrop: No, that is exactly the point. The contract of the method says give me a runnable, I will call it, and if it throws an exception, I will ensure that you will see an IOException. That's why I check both directions. I agree, that _I will call it_, is quite expected, but it is still part of the contract to be tested.

whiskeysierra: If nobody would call it there would be no exception. And since we're checking for the exception already...

jhorstmann: > If nobody would call it there would be no exception

Except if the implementation of `ensureIOException` was just `throw new IOException()`.

whiskeysierra: Not true, as that wouldn't match the expected exception stated in the test. With the same idea in mind one could also easily fool the current version of the test by doing this:

```java
try {
    runnable.run();
} catch (Exception e) {
    throw new IOException(); // not using the right one
}
```

The underlying problem is that the exception verification is not strict enough. Instead of just comparing the types it should probably be based on identity. The following would be a more correct version of that test that would do things only once while being more concise.

```java
@Test
public void runnableShouldCastException() throws Exception {
    final Exception e = new Exception();
    exception.expect(IOException.class);
    exception.expectCause(is(e));

    final ThrowingRunnable runnable = mock(ThrowingRunnable.class);

    doThrow(e).when(runnable).run();
    ensureIOException(runnable);
}
```

tkrop: Yes, I agree. I will change it to that.

riptide-stream/src/test/java/org/zalando/riptide/stream/StreamsTest.java
```diff
@@ -211,7 +211,7 @@ public void shouldCallConsumerWithoutStream() throws Exception {
 
         verify(verifier).accept(
                 new AccountBody("1234567890", "Acme Corporation"));
-        verify(verifier, times(1)).accept(any());
+        verify(verifier).accept(any());
```
whiskeysierra: Should probably be `not(eq(new AccountBody("1234567890", "Acme Corporation")))`

whiskeysierra:
:+1:

tkrop:
:+1:

PR #150, opened 2016-08-23
Simplified exception handling for clients

### LINKED ISSUE #143: Replace IOException with UncheckedIOException?
@jhorstmann @lukasniemeier-zalando

whiskeysierra:
Fixes #143

### DIFF
--- a/riptide-core/src/test/java/org/zalando/riptide/IOTest.java
+++ b/riptide-core/src/test/java/org/zalando/riptide/IOTest.java
@@ -28,11 +28,15 @@ import org.junit.Test;
 
 import java.io.IOException;
 import java.util.List;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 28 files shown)


riptide-core/src/test/java/org/zalando/riptide/IOTest.java
```diff
@@ -92,10 +96,23 @@ public void shouldReadContributors() throws IOException, ExecutionException, Int
...
+
+        future.cancel(true);
+
+        try {
+            future.join();
+        } catch (final CancellationException e) {
+            // expected
+        }
+
+        driver.reset();
```
tkrop: Shouldn't this be in `after()`?

tkrop:
:+1:

riptide-core/src/test/java/org/zalando/riptide/IOTest.java
whiskeysierra: No. There is already a `server.verify()` in the teardown and we want to skip that because we're issuing a request that we want to cancel.

tkrop: I can only see and `executor.shutdown()` ...

Anyhow, why do you need to reset at all? According to the theory you mentioned last week, it should run in an isolated test instance.

PR #155, opened 2016-08-29
added import for fauxpas

tkrop:


pom.xml
```diff
@@ -100,6 +101,11 @@
 
             <!-- common -->
             <dependency>
+	        <groupId>org.zalando</groupId>
+		<artifactId>faux-pas</artifactId>
+		<version>${fauxpas.version}</version>
```
whiskeysierra: We should drop the version from `riptide-core/pom.xml`

tkrop:
:+1:

PR #165, opened 2016-09-14
Changed query param support from String values to Object

### LINKED ISSUE #162: Support queryParam(String, Object)
and call toString on the value

whiskeysierra:
Fixes #162

### DIFF
--- a/riptide-core/src/test/java/org/zalando/riptide/RequesterTest.java
+++ b/riptide-core/src/test/java/org/zalando/riptide/RequesterTest.java
@@ -88,14 +88,15 @@ public class RequesterTest {
 
     @Test
     public void shouldAppendQueryParams() {
-        server.expect(requestTo("https://api.example.com?foo=bar&foo=baz&bar=null"))
+        server.expect(requestTo("https://api.example.com?foo=true&foo=null&foo=bar&bar=123"))
                 .andRespond(withSuccess());
 
         unit.head("https://api.example.com")
-                .queryParam("foo", "bar")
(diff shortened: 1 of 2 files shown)


riptide-core/src/test/java/org/zalando/riptide/RequesterTest.java
```diff
@@ -88,14 +88,15 @@ public void shouldEncodeInlinedQueryParams() {
 
     @Test
     public void shouldAppendQueryParams() {
-        server.expect(requestTo("https://api.example.com?foo=bar&foo=baz&bar=null"))
+        server.expect(requestTo("https://api.example.com?foo=true&foo=null&foo=bar&bar=123"))
```
jhorstmann: Serializing `null` as string `"null"` might be unexpected, but seems this was already done the same way before this change.

whiskeysierra: Indeed, that was the same before. What would you expect? An empty string?

whiskeysierra: With quotes? oO

lukasniemeier-zalando: > With quotes? oO

No without :see_no_evil: but whatever we do here is probably unexpected from someone.

lukasniemeier-zalando:
We've discussed this issue again and we will stick with `String` for now to prevent any API confusion.

PR #166, opened 2016-09-14
Fixed incorrect consumption when bubbling up while dispatching

### LINKED ISSUE #161: NoRouteException should not consume anything while bubbling up
Given the following routing tree

```java
rest.post("/event-types/{event-type}/events", "customs.tour-change")
        .body(singleton(event))
        .dispatch(series(),
                on(SUCCESSFUL).dispatch(status(),
                        on(HttpStatus.OK).call(() -> capture.accept(emptyList())),
                        on(HttpStatus.MULTI_STATUS).call(listOf(Response.class), capture)),
                on(CLIENT_ERROR).dispatch(status(),
                        on(HttpStatus.UNPROCESSABLE_ENTITY).call(listOf(Response.class), capture)),
                anySeries().call(problemHandling()))
        .thenApply(capture)
        .thenAccept(this::failIfNecessary)
        .join();
```

if a `400 Bad Request` is returned the stream will be consumed and then it can't be processed by the wildcard match on the top level.

whiskeysierra:
Fixes #161

### DIFF
--- a/riptide-core/src/main/java/org/zalando/riptide/DefaultRoutingTree.java
+++ b/riptide-core/src/main/java/org/zalando/riptide/DefaultRoutingTree.java
@@ -1,9 +1,7 @@
 package org.zalando.riptide;
 
 import org.springframework.http.client.ClientHttpResponse;
-import org.zalando.fauxpas.ThrowingSupplier;
 
-import java.io.IOException;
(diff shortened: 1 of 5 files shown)


lukasniemeier-zalando:
:+1:

PR #169, opened 2016-09-21
Cleaned up handling of URI Templates and URIs in regards to transfomation

### LINKED ISSUE #168: Resolve URIs against Base URL
URIs are currently taken as-is, i.e. base URL is not prepended. This leads to weird scenarios where clients want to use URI facilities, e.g. resolve, but still need the base URL.

- [x] add example table to README
- [ ] describe URL transformation
- [ ] release notes describing the breaking change (https://example.org/api + /users is **not** https://example.org/api/users but rather https://example.org/users)

whiskeysierra:
Cleaned up handling of URI Templates and URIs in regards to transformation

This includes but is not limited to
- parameter expansion
- encoding
- resolving against base url
- query string
- normalization

Fixes #168


whiskeysierra:
@lukasniemeier-zalando Please verify

PR #172, opened 2016-10-11
Dropped Completion in favor of CompletableFuture

### LINKED ISSUE #163: Remove Completion in favor of CompletableFuture
It doesn't provide much value over CompletableFuture, in fact for most use cases you need to transform it anyway, e.g. to use `CompletableFuture.allOf()` etc.

whiskeysierra:
Fixes #163


lukasniemeier-zalando:
:+1:

PR #178, opened 2016-11-21
Keep original stack trace

### LINKED ISSUE #177: Stacktraces don't show original caller
Since we hand over the request execution to a thread pool already, we loose the original call site if an exception occurs. Makes it a lot harder to debug an issue than it should be.

whiskeysierra:
Fixes #177

### DIFF
--- a/riptide-core/src/main/java/org/zalando/riptide/NoWildcardException.java
+++ b/riptide-core/src/main/java/org/zalando/riptide/NoWildcardException.java
@@ -2,4 +2,13 @@ package org.zalando.riptide;
 
 public final class NoWildcardException extends RuntimeException {
 
+    /**
+     * We don't care for the stack trace, and since it's expensive to calculate we just use the same instance
+     * all the time. It's exclusively used to jump around the stack.
+     */
+    @Override
+    public Throwable fillInStackTrace() {
+        return this;
+    }
+
 }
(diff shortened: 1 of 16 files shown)


riptide-core/src/main/java/org/zalando/riptide/NoWildcardException.java
```diff
@@ -2,4 +2,13 @@
 
 public final class NoWildcardException extends RuntimeException {
 
+    /**
+     * We don't care for the stack trace, and since it's expensive to calculate we just use the same instance
+     * all the time. It's exclusively used to jump around the stack.
```
lukasniemeier-zalando: `we just use the same instance all the time` is not true anymore as it seems

riptide-core/src/main/java/org/zalando/riptide/Requester.java
```diff
@@ -83,36 +87,9 @@ public final Requester contentType(final MediaType contentType) {
...
-                        future.completeExceptionally(e);
-                    }
-                }, future::completeExceptionally);
-
-                return future;
-            }
+        final Exception exceptionWithOriginalStackTrace = new Exception();
 
-        };
+        return new ResponseDispatcher(worker, listenable, exceptionWithOriginalStackTrace);
```
lukasniemeier-zalando: Why not pass only `new Throwable().getStackTrace()` here?

riptide-core/src/main/java/org/zalando/riptide/NoWildcardException.java
whiskeysierra: Correct, fixed.

lukasniemeier-zalando:
:+1:

PR #179, opened 2016-11-28
Added static join method that rethrows the cause of a completion exception

whiskeysierra:
Requires #178 to be merged first


lukasniemeier-zalando:
:+1:

PR #181, opened 2016-12-01
Added riptide-problem and riptide-exceptions

whiskeysierra:


whiskeysierra:
> Cool stuff, should we add a small paragraph each to the README? Maybe even a table is sufficient.

Absolutely. I'm not sure whether each module deserves its own README or just a section in the main. I'd like the keep the big one as small as possible.

riptide-exceptions/src/main/java/org/zalando/riptide/exceptions/TemporaryExceptionPlugin.java
```diff
@@ -0,0 +1,66 @@
...
+
+import static org.zalando.fauxpas.FauxPas.rethrow;
+import static org.zalando.fauxpas.FauxPas.throwingFunction;
+
+public final class TemporaryExceptionPlugin implements Plugin {
+
+    private final Predicate<Throwable> temporary;
+
+    public TemporaryExceptionPlugin() {
+        this(InterruptedIOException.class::isInstance,
```
jhorstmann: IOExceptions in general are not retryable?

whiskeysierra: No, at least not by default. There is just too much stuff that is an IOException that is a permanent error:

- CharacterCodingException
- CharConversionException
- FileNotFoundException
- InvalidPropertiesFormatException
- MalformedURLException
- UnsupportedEncodingException

To name a few (just from the JDK alone).

lukasniemeier-zalando:
:+1:

PR #183, opened 2017-03-24
Added support for problemHandling(ThrowingConsumer)

whiskeysierra:


riptide-problem/src/main/java/org/zalando/riptide/problem/ProblemRoute.java
```diff
@@ -10,17 +14,43 @@
 
 public final class ProblemRoute {
 
-    private static final Route ROUTE = dispatch(contentType(),
-            on(parseMediaType("application/problem+json")).call(Exceptional.class, Exceptional::propagate),
-            on(parseMediaType("application/x.problem+json")).call(Exceptional.class, Exceptional::propagate),
-            on(parseMediaType("application/x-problem+json")).call(Exceptional.class, Exceptional::propagate));
+    private static final MediaType PROBLEM = parseMediaType("application/problem+json");
```
AlexanderYastrebov: So you may just hardcode them into `problemHandling(final ThrowingConsumer<Exceptional, ? extends Exception> consumer)`

riptide-problem/src/test/java/org/zalando/riptide/problem/ProblemRouteTest.java
```diff
@@ -70,4 +81,20 @@ private void perform(final String mediaType) {
...
+                        .body(new ClassPathResource("problem.json"))
+                        .contentType(MediaType.parseMediaType("application/problem+json")));
+
+        unit.get(url)
+                .dispatch(series(),
+                        on(SUCCESSFUL).call(pass()),
+                        anySeries().call(problemHandling(consumer)))
+                .join();
+
+        verify(consumer).tryAccept(any());
```
AlexanderYastrebov: You may also capture and assert on what was accepted

whiskeysierra:
Then I'd parse them over and over again.

AlexanderYastrebov:
:+1:

riptide-problem/src/main/java/org/zalando/riptide/problem/ProblemRoute.java
```diff
@@ -10,17 +14,43 @@
...
+    private static final MediaType X_DOT_PROBLEM = parseMediaType("application/x-problem+json");
+    private static final MediaType X_DASH_PROBLEM = parseMediaType("application/x.problem+json");
```
jhorstmann: The two names here are switched

whiskeysierra: Good catch. Fixed.

riptide-problem/src/main/java/org/zalando/riptide/problem/ProblemRoute.java
```diff
@@ -10,17 +14,51 @@
...
+     * Produces a {@link Route route} that dispatches on the {@link Navigators#contentType() content type} and
+     * recognises {@code application/problem+json} as well as {@code application/x-problem+json} and
+     * {@code application/x.problem+json} as {@link Problem problems} and handles them given the supplied consumer.
+     *
+     * @param consumer the exception handler
+     * @return a route for handling problems dynamically
+     */
+    public static Route problemHandling(final ThrowingConsumer<Exceptional, ? extends Exception> consumer) {
+        return dispatch(contentType(),
+                on(PROBLEM).call(Exceptional.class, consumer),
```
AlexanderYastrebov: BTW what about making `on(T... values)`?

whiskeysierra: I prefer individual bindings per value. I changed the implementation to show how to deal with a situation like that. It's a bit more verbose, but I think it's also more readable/understandable.

AlexanderYastrebov:
:+1:

PR #184, opened 2017-03-24
Made Rest.execute(HttpMethod, ..) public

whiskeysierra:
This allows for use cases where the http method is dynamically determined at runtime. Right now that requires a lookup/mapping from http method to function, e.g.

```java
switch (method) {
    case HEAD:
        return rest.head(uri);
    ...
}
```


AlexanderYastrebov:
:+1:

PR #185, opened 2017-03-25
Cleaned up dependencies

whiskeysierra:


riptide-httpclient/pom.xml
```diff
@@ -21,8 +21,15 @@
 
     <dependencies>
         <dependency>
-            <groupId>org.zalando</groupId>
-            <artifactId>riptide-core</artifactId>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-web</artifactId>
```
AlexanderYastrebov: Is it possible to use riptide-httpclient without `core` by itself?

whiskeysierra: Yes. There is no code dependency. The pom correctly reflects that now.

AlexanderYastrebov:
:+1:

PR #188, opened 2017-03-27
Made getCharset backwards compatible

whiskeysierra:


AlexanderYastrebov:
I will cross-post the discussion so it could be traced back.

If I get it right they introduced alias and deprecated existing one http://docs.spring.io/spring-framework/docs/4.3.x/javadoc-api/org/springframework/util/MimeType.html#getCharset--

It seems there is no urgent need in Riptide to use new one in favour of deprecated.
We may stick to old one in Riptide and then switch to new one only when Riptide's Spring dependency will be updated and deprecated method removed.

Using the old one allows to use Riptide with outdated spring versions, provided they are interface compatible.

AlexanderYastrebov:
:+1:

PR #192, opened 2017-03-28
Added Spring 4.1.0 till 4.3.7 support

whiskeysierra:


riptide-core/src/main/java/org/zalando/riptide/RestBuilder.java
```diff
@@ -1,17 +1,24 @@
...
 import org.springframework.web.client.RestTemplate;
 
 import javax.annotation.Nullable;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.RejectedExecutionException;
```
AlexanderYastrebov: Why these imports were added?

riptide-core/src/test/java/org/zalando/riptide/FailedDispatchTest.java
```diff
@@ -139,9 +137,12 @@ public void shouldThrowOnFailedConversionBecauseOfFaultyBody() {
 
     @Test
     public void shouldHandleNoBodyAtAll() {
+        final HttpHeaders headers = new HttpHeaders();
+        headers.setContentLength(0);
```
AlexanderYastrebov: Will it work without content-length header?

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamConverter.java
```diff
@@ -32,15 +31,26 @@
             Collections.unmodifiableList(Arrays.asList(APPLICATION_JSON_SEQ, APPLICATION_X_JSON_STREAM));
 
     private final ObjectMapper mapper;
-    private final List<MediaType> medias;
+    private final List<MediaType> supportedMediaTypes;
 
+    /**
+     * @deprecated Use {@link Streams#streamConverter(ObjectMapper)}
+     */
+    @Deprecated
```
AlexanderYastrebov: If it is deprecated, what is the alternative?

AlexanderYastrebov: Nevermind

spring-compatibility-test.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh -e
+
+grep -oP "SPRING_VERSION=\K(.+)" .travis.yml | tac | while read version; do
+    logfile=$(mktemp)
+    echo ${version}: ${logfile}
+    ./mvnw clean test-compile -D spring.version=${version} -l ${logfile}
+done
```
AlexanderYastrebov: \n

README.md
```diff
@@ -116,7 +116,9 @@ on(SUCCESSFUL).dispatch(contentType(),
 | `ThrowingConsumer<T>`                  | `on(..).call(TypeToken<T>, ThrowingConsumer<T>)`    |
 | `RoutingTree`                          | `on(..).dispatch(..)`                               |
 
-> 
+## Dependencies
+
+- Spring 4.1 or higher
```
AlexanderYastrebov: I would declare full version of 4.1.0

pom.xml
```diff
@@ -72,7 +72,7 @@
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
         <slf4j.version>1.7.21</slf4j.version>
-        <spring.version>4.3.2.RELEASE</spring.version>
+        <spring.version>4.1.0.RELEASE</spring.version>
```
AlexanderYastrebov: But here we may stick with latest spring version because you have implemented backward compatibility, not forward

.travis.yml
```diff
@@ -9,7 +9,38 @@ cache:
   directories:
     - $HOME/.m2
 
-script: ./mvnw clean verify
+env:
+  matrix:
+    - SPRING_VERSION=4.1.0.RELEASE
```
jhorstmann: Is it really necessary to build with all these versions, doesn't that slow down testing a lot? I would expect the patch releases to keep the api compatible.

riptide-core/src/test/java/org/zalando/riptide/FailedDispatchTest.java
whiskeysierra: Spring changed its heuristic to detect an empty body somewhere between 4.1 and 4.3. Before it checked for 204 No Content and the Content-Length. In the later versions it also peeks at the first byte of the body.

riptide-core/src/main/java/org/zalando/riptide/RestBuilder.java
whiskeysierra: For 4.0.x I'd have to implement a custom executor. Those are relicts of that test. I'll remove it.

.travis.yml
AlexanderYastrebov: Lets see and revert if it creates a problem

riptide-stream/src/main/java/org/zalando/riptide/stream/StreamConverter.java
whiskeysierra: There shouldn't be a possibility to not pass in a mapper. Clients should always pass one.

spring-compatibility-test.sh
whiskeysierra: Pffff

pom.xml
whiskeysierra: No idea why this made it into the commit. I'll change it.

.travis.yml
whiskeysierra: > Is it really necessary to build with all these versions

I did it today and a lot of them introduced issues.

> doesn't that slow down testing a lot

Travis should execute them in parallel.

> would expect the patch releases to keep the api compatible.

API yes (mostly, at least) but the implementation sometimes changes in an imcompatible way.

AlexanderYastrebov:
:+1:

Issue #193, opened 2017-04-07
Add support for dynamic base url


AlexanderYastrebov:
As a developer
I want to have support for dynamic base url
So that it is possible to change baseUrl after Rest instance is constructed

Main usecase scenario is to use dynamic config provider (aka config service) to get base url which may change over time/after config update.

PR #194, opened 2017-04-07
GH-193 Adds support for dynamic base url

AlexanderYastrebov:
Fixes GH-193

This PR adds support for dynamic base URL.

`RequestArguments` interface was cleaned from `withX` builder methods. The rationale is that single responsibility of this interface is to represent view over request arguments. Since interface is public it is a breaking change. The impact should be discussed and change reverted if **really** necessary.

Decision should be made about providing accompanying `RestBuilder::baseUrlProvider(Supplier<String> baseUrlProvider)` method since base url often comes in form of a string value as well.

### DIFF
--- a/riptide-core/src/main/java/org/zalando/riptide/DefaultRequestArguments.java
+++ b/riptide-core/src/main/java/org/zalando/riptide/DefaultRequestArguments.java
@@ -4,6 +4,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMultimap;
 import java.net.URI;
 import java.util.Optional;
+import java.util.function.Supplier;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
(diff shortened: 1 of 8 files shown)


riptide-core/src/main/java/org/zalando/riptide/DefaultRequestArguments.java
```diff
@@ -21,7 +27,7 @@
     HttpMethod method;
 
     @Wither
-    URI baseUrl;
+    Supplier<URI> baseUrlProvider;
```
AlexanderYastrebov: This is also a breaking change if `DefaultRequestArguments` is really considered public and not for library-internal use only

AlexanderYastrebov: Though this could be solved by providing `withBaseUrl` method

AlexanderYastrebov:
After some thought I do not like that RequestArguments delegate call for base url to provider. This may lead to situation when actual request url and url observed e.g. by plugin are different. Better way is to freeze baseUrl during request preparation.

AlexanderYastrebov:
Closing in favour of https://github.com/zalando/riptide/pull/195

PR #195, opened 2017-04-07
GH-193 Adds support for dynamic base url

### LINKED ISSUE #193: Add support for dynamic base url

AlexanderYastrebov:
Fixes #193

### DIFF
--- a/riptide-core/src/main/java/org/zalando/riptide/Rest.java
+++ b/riptide-core/src/main/java/org/zalando/riptide/Rest.java
@@ -23,7 +23,7 @@ public final class Rest {
             final Supplier<URI> baseUrlProvider, final Plugin plugin) {
         this.requestFactory = checkNotNull(requestFactory, "request factory");
         this.worker = new MessageWorker(converters);
-        this.baseUrlProvider = checkNotNull(baseUrlProvider, "baseUrl provider");
+        this.baseUrlProvider = checkNotNull(baseUrlProvider, "base url provider");
         this.arguments = RequestArguments.create();
         this.plugin = plugin;
     }
--- a/riptide-core/src/main/java/org/zalando/riptide/Rest.java
+++ b/riptide-core/src/main/java/org/zalando/riptide/Rest.java
@@ -5,9 +5,9 @@ import org.springframework.http.HttpMethod;
 import org.springframework.http.client.AsyncClientHttpRequestFactory;
 import org.springframework.http.converter.HttpMessageConverter;
 
-import javax.annotation.Nullable;
 import java.net.URI;
 import java.util.List;
+import java.util.function.Supplier;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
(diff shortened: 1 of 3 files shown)


jhorstmann:
:+1:

riptide-core/src/main/java/org/zalando/riptide/RestBuilder.java
```diff
@@ -59,7 +60,11 @@ public RestBuilder baseUrl(@Nullable final String baseUrl) {
     }
 
     public RestBuilder baseUrl(@Nullable final URI baseUrl) {
-        this.baseUrl = baseUrl;
+        return baseUrlProvider(() -> baseUrl);
+    }
+
+    public RestBuilder baseUrlProvider(final Supplier<URI> baseUrlProvider) {
```
AlexanderYastrebov: This method could be named `baseUrl` as well

riptide-core/src/main/java/org/zalando/riptide/Rest.java
```diff
@@ -5,24 +5,26 @@
...
     private final RequestArguments arguments;
     private final Plugin plugin;
 
     Rest(final AsyncClientHttpRequestFactory requestFactory, final List<HttpMessageConverter<?>> converters,
-            @Nullable final URI baseUrl, final Plugin plugin) {
+            final Supplier<URI> baseUrlProvider, final Plugin plugin) {
         this.requestFactory = checkNotNull(requestFactory, "request factory");
         this.worker = new MessageWorker(converters);
-        this.arguments = RequestArguments.create().withBaseUrl(baseUrl);
+        this.baseUrlProvider = checkNotNull(baseUrlProvider, "baseUrl provider");
```
whiskeysierra: `"base url provider"`

riptide-core/src/main/java/org/zalando/riptide/RestBuilder.java
whiskeysierra: Yes, please.

riptide-core/src/test/java/org/zalando/riptide/DynamicBaseUrlTest.java
```diff
@@ -0,0 +1,72 @@
...
+import static org.zalando.riptide.Navigators.series;
+import static org.zalando.riptide.Route.pass;
+
+public class DynamicBaseUrlTest {
+
+    private final Rest unit;
+    private final MockRestServiceServer server;
+
+    @SuppressWarnings("unchecked")
+    private final Supplier<URI> baseUrlProviderMock = mock(Supplier.class);
```
whiskeysierra: I'd just use an `AtomicReference` as `baseUrl::get`. It's easier to use in your test.

riptide-core/src/test/java/org/zalando/riptide/DynamicBaseUrlTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Before
+    @SuppressWarnings("unchecked")
+    public void before() {
+        reset(baseUrlProviderMock);
+    }
+
+    @After
+    public void after() {
+        server.verify();
```
whiskeysierra: I'd move this into the test.

riptide-core/src/test/java/org/zalando/riptide/DynamicBaseUrlTest.java
```diff
@@ -0,0 +1,72 @@
...
+        final MockSetup setup = new MockSetup();
+
+        this.unit = setup.getRestBuilder().baseUrlProvider(baseUrlProviderMock).build();
+        this.server = setup.getServer();
+    }
+
+    @Before
+    @SuppressWarnings("unchecked")
+    public void before() {
+        reset(baseUrlProviderMock);
```
whiskeysierra: Not needed.

riptide-core/src/test/java/org/zalando/riptide/DynamicBaseUrlTest.java
AlexanderYastrebov: I want to verify number of calls to it as well

whiskeysierra: Is that really important for your test?`If yes, I'd suggest to use a second test for that.

whiskeysierra:
👍

PR #196, opened 2017-04-10
GH-189 Adds Rest methods to query baseUrl

### LINKED ISSUE #189: Provide support for querying configured baseUrl
As a developer
I want to query Rest's `baseUrl` without explicitly providing any path
So that it it possible to do so

Proposal is to add no-argument get, post, put, etc methods which will call preconfigured baseUrl or fail if there is no one configured.

AlexanderYastrebov:
Fixes #189

### DIFF
--- a/riptide-core/src/main/java/org/zalando/riptide/RequestArguments.java
+++ b/riptide-core/src/main/java/org/zalando/riptide/RequestArguments.java
@@ -8,13 +8,13 @@ import org.springframework.web.util.UriComponentsBuilder;
 
 import javax.annotation.Nullable;
 import java.net.URI;
-import java.util.Optional;
 
+import static com.google.common.base.Preconditions.checkNotNull;
 import static org.springframework.web.util.UriComponentsBuilder.fromUri;
 import static org.springframework.web.util.UriComponentsBuilder.fromUriString;
 
 public interface RequestArguments {
-    
+
     URI getBaseUrl();
 
     HttpMethod getMethod();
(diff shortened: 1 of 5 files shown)


riptide-core/src/main/java/org/zalando/riptide/RequestArguments.java
```diff
@@ -52,15 +52,34 @@
     RequestArguments withBody(@Nullable Object body);
 
     default RequestArguments withRequestUri() {
-        // expand uri template
-        final URI uri = Optional.ofNullable(getUri())
-                .orElseGet(() -> fromUriString(getUriTemplate())
+        final URI uri = getUri();
+
+        final URI toResolve;
+        if (uri != null) {
```
whiskeysierra: I'd invert both ifs.

riptide-core/src/main/java/org/zalando/riptide/RequestArguments.java
```diff
@@ -52,15 +52,34 @@
...
+            }
+        }
 
-        // resolve uri against base url
-        final URI resolved = getBaseUrl() == null ? uri : getBaseUrl().resolve(uri);
+        final URI baseUrl = getBaseUrl();
+
+        final URI resolved;
+        if (toResolve == null) {
+            resolved = checkNotNull(baseUrl, "base url required");
```
whiskeysierra: Illegal argument or illegal state would also fit.

AlexanderYastrebov:
It is not the first time I see failed build due ti `TemporaryExceptionPluginTest.shouldClassifyAsPermanent`
Last time I just closed/reopened PR to retrigger build and it had gone

riptide-core/src/main/java/org/zalando/riptide/RequestArguments.java
AlexanderYastrebov: :)

riptide-core/src/main/java/org/zalando/riptide/RequestArguments.java
AlexanderYastrebov: I decided to use the same Preconditions.checkNotNull as everywhere else.

AlexanderYastrebov:
Raised issue regarding of build failure #197

whiskeysierra:
Much better now :stuck_out_tongue_winking_eye:

AlexanderYastrebov:
:+1:

PR #198, opened 2017-04-26
Attempt to fix race condition in test

### LINKED ISSUE #197: Build sometimes fails due to TemporaryExceptionPluginTest.shouldClassifyAsPermanent
E.g. https://travis-ci.org/zalando/riptide/builds/220564959

I think it is because `TemporaryExceptionPluginTest` configures client socket timeout to 500 ms and it sometimes happens before test method `shouldClassifyAsPermanent` works out

whiskeysierra:
Fixes #197 (hopefully)


AlexanderYastrebov:
:+1:

PR #200, opened 2017-04-27
Added support for non-standard url resolution

whiskeysierra:


riptide-core/src/main/java/org/zalando/riptide/RestBuilder.java
```diff
@@ -79,6 +80,11 @@ private URI checkAbsoluteBaseUrl(@Nullable final URI baseUrl) {
         return baseUrl;
     }
 
+    public RestBuilder urlResolution(@Nullable final UrlResolution resolution) {
```
AlexanderYastrebov: Does it make sense to allow nulls here or would it be better to check and fail?

whiskeysierra: The builder allows nulls basically everywhere, I wanted to be consistent. It's nice so clients can read those values directly from their configuration and don't have to double check.

riptide-core/src/main/java/org/zalando/riptide/UrlResolution.java
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Resolves a given relative URI by appending its path to and merging its query parameters with the given base URL.
+     *
+     * @see UriComponentsBuilder#pathSegment(String...)
+     */
+    APPEND {
+        @Override
+        URI resolve(final URI baseUrl, final URI uri) {
+            return UriComponentsBuilder.fromUri(baseUrl)
+                    .pathSegment(uri.getRawPath())
```
AlexanderYastrebov: Should it be `.path(...)`

http://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/util/UriComponentsBuilder.html#pathSegment-java.lang.String...-
> Each path segment may contain URI template variables and should not contain any slashes

whiskeysierra: I tried `path(..)` but it won't work with the following:

Base URL: http://example.org/api
URI: foo

Will produce: **http://example.org/apifoo**

AlexanderYastrebov: That API comment looks suspicious, I think we miss test cases with multi-segment request paths like `http://example.com` + `/first/second`

whiskeysierra: That would work:

```java
final UriComponentsBuilder builder = UriComponentsBuilder.fromUri(baseUrl);

@Nullable final String path = uri.getRawPath();

if (!nullToEmpty(path).isEmpty()) {
    builder.path("/").path(path);
}

return builder
        .replaceQuery(uri.getRawQuery())
        .build()
        .toUri();
```

whiskeysierra: > That API comment looks suspicious, I think we miss test cases with multi-segment request paths like http://example.com + /first/second

I just added those test cases. But it actually works with `pathSegment`.

riptide-core/src/main/java/org/zalando/riptide/UrlResolution.java
```diff
@@ -0,0 +1,58 @@
...
+        @Override
+        URI resolve(final URI baseUrl, final URI uri) {
+            return baseUrl.resolve(uri);
+        }
+    },
+
+    /**
+     * Resolves a given relative URI by appending its path to and merging its query parameters with the given base URL.
+     *
+     * @see UriComponentsBuilder#pathSegment(String...)
```
AlexanderYastrebov: This one is not used anymore, isn't it?
Also I am still thinking of cornercases like `http://example.com` + `/` and `http://example.com/` + `/` :)

whiskeysierra: Done

AlexanderYastrebov:
:+1:

PR #202, opened 2017-04-27
Added first draft of graphviz module

### LINKED ISSUE #187: Render routing tree
A graphviz module would be an interesting way to see how useful the introspection api of routes and routing trees actually is.

whiskeysierra:
Fixes #187 

- [ ] README


riptide-core/src/main/java/org/zalando/riptide/Route.java
```diff
@@ -115,4 +118,39 @@ static Route call(final ThrowingConsumer<ClientHttpResponse, ? extends Exception
...
+    static Route named(final Object name, final Route route) {
+        return new Route() {
+            @Override
+            public void execute(final ClientHttpResponse response, final MessageReader reader) throws Exception {
+                route.execute(response, reader);
+            }
+
+            @Override
+            public String toString() {
+                final ThrowingSupplier<String, Exception> stringer = () -> {
```
AlexanderYastrebov: Why lambda is needed?

riptide-core/src/main/java/org/zalando/riptide/Route.java
```diff
@@ -115,4 +118,39 @@ static Route call(final ThrowingConsumer<ClientHttpResponse, ? extends Exception
...
+                    final Method toString = name.getClass().getMethod("toString");
+                    return toString.getDeclaringClass() == Object.class ? "<code>" : name.toString();
```
AlexanderYastrebov: 1. getDeclaringClass may return null
2. Why is it called on `toString` Method instead of `name` directly?

whiskeysierra: 1. For `toString` that's impssible, but even then, since we use `==` it's null safe.
2. I want to reflect changes to the underlying object's tostring representation. If I call it early we're caching it. Probably not a big deal, but the way it's done right now should have the least surprises.

riptide-core/src/main/java/org/zalando/riptide/Route.java
whiskeysierra: Reflection throws checked exceptions but I can't rethrow them and `SneakyThrows` is not 100% coverable. Open for other suggestions.

AlexanderYastrebov: old school try-catch?

whiskeysierra: There is no way to force those exceptions since `toString` always exists and is always accessible.

AlexanderYastrebov: Feels magic, I do not get it. Do you mean that it is impossible to cover it with test?

whiskeysierra: Yes, there is no way to provoke those exceptions and therefore you can' test it.

whiskeysierra: It's worth to put `@Hack` there, at least.

whiskeysierra:
I decided to keep this module as a separate (personal) repo outside of riptide since it's more of a *proof of concept* rather than something useful for a majority of clients.

PR #206, opened 2017-06-20
Added support for fallback problem route

### LINKED ISSUE #205: question // "orElse" route
The code I have until now in my application is the following (in `Kotlin`, I hope it is still understandable).

```kotlin
            getDispatcher(http)
                .dispatch(Navigators.series(),
                    Bindings.on(HttpStatus.Series.SUCCESSFUL).call { response ->
                        deferred.resolve(Response(response, extract(response, clazz)))
                    },
                    Bindings.anySeries().dispatch(
                        Navigators.status(),
                        *errorBindings,
                        Bindings.anyStatus().call(ProblemRoute.problemHandling {
                            deferred.reject(it as Exception)
                        })
                    )
                )
                .join()
```

The problem I have here is that if the error that is returned is not of one of the content-type managed by `ProblemRoute.problemHandling`, the complete call fails with a Riptide exception about no route being defined for this case.

I am looking for a way to add something after the use of `problemHandling` that would semantically means: "If at this point, there was no route defined for this case, do <this>".

For reference, here is the code for `problemHandling`:
```java
        final Route route = Route.call(Exceptional.class, consumer);
        return dispatch(contentType(),
                on(PROBLEM).call(route),
                on(X_DOT_PROBLEM).call(route),
                on(X_DASH_PROBLEM).call(route));
```

The way I understand it, I would need to insert myself as a last parameter inside of that call to `dispatch` (after `on(X_DASH_PROBLEM.call(route)`).

To make my case work, I had to import some code from `ProblemRoute` into my own application and do something like this:
```kotlin
// some static stuff
        private val PROBLEM = parseMediaType("application/problem+json")
        private val X_DOT_PROBLEM = parseMediaType("application/x.problem+json")
        private val X_DASH_PROBLEM = parseMediaType("application/x-problem+json")
// and then, in my method
            val problemJsonRoute = Route.call(Exceptional::class.java) {
                deferred.reject(it as Exception)
            }
            val orElseConsumer: (ClientHttpResponse) -> Unit = {
                deferred.reject(Exception(it.body.bufferedReader().readText()))
            }
            getDispatcher(http)
                .dispatch(
                    Navigators.series(),
                    Bindings.on(SUCCESSFUL).call { response ->
                        deferred.resolve(Response(response, extract(response, clazz)))
                    },
                    Bindings.anySeries().call(RoutingTree.dispatch(
                        Navigators.status(),
                        *errorBindings,
                        Bindings.anyStatus().call(dispatch<MediaType>(
                            contentType(),
                            on<MediaType>(PROBLEM).call(problemJsonRoute),
                            on<MediaType>(X_DOT_PROBLEM).call(problemJsonRoute),
                            on<MediaType>(X_DASH_PROBLEM).call(problemJsonRoute),
                            Bindings.anyContentType().call(orElseConsumer)
                        ))
                    ))
                )
                .join()
```

It works and I could just go on like this, but I feel that is a shame to not use existing library code (`ProblemRoute`).

Is there a "clean" way of doing what I want?
I am quite new to _Riptide_ and might not understand correctly the dispatching/routing system.

Thanks for any assistance you can provide!

whiskeysierra:
Fixes #205

### DIFF
--- a/riptide-problem/README.md
+++ b/riptide-problem/README.md
@@ -82,6 +82,10 @@ of Riptide and fail with a `NoRouteException` (unless a wildcard matches). This 
 http.post("/").dispatch(series(),
     on(SUCCESSFUL).call(pass()),
     anySeries().call(problemHandling(call(this::onUnsupportedError))));
+
+void onUnsupportedError(ClientHttpResponse response) throws IOException {
+    // TODO handle non-problem error response here
+}
 ```
 
 ## Getting Help
--- a/riptide-problem/README.md
+++ b/riptide-problem/README.md
@@ -63,6 +63,27 @@ try {
 }
 ```
 
+### Custom Handling
+
+If throwing a problem is not the desired behaviour one can override it by passing in a custom consumer:
+ 
+```java
+http.post("/").dispatch(series(),
+    on(SUCCESSFUL).call(pass()),
+    anySeries().call(problemHandling(e -> LOG.error("Unexpected problem", e))));
+```
+
+### Fallback Route
+
+If the `ProblemRoute` fails to dispatch, e.g. because of a different media type, it will follow the default behaviour
+of Riptide and fail with a `NoRouteException` (unless a wildcard matches). This behaviour can be overridden by:
+
+```java
+http.post("/").dispatch(series(),
+    on(SUCCESSFUL).call(pass()),
+    anySeries().call(problemHandling(call(this::onUnsupportedError))));
+```
+
 ## Getting Help
 
 If you have questions, concerns, bug reports, etc., please file an issue in this repository's [Issue Tracker](../../../../issues).
--- a/riptide-problem/src/main/java/org/zalando/riptide/problem/ProblemRoute.java
+++ b/riptide-problem/src/main/java/org/zalando/riptide/problem/ProblemRoute.java
@@ -80,6 +80,7 @@ public final class ProblemRoute {
      * {@code application/x.problem+json} as {@link Problem problems} and handles them given the supplied consumer.
      * The given fallback will be used if none of the mentioned content types matches.
      *
+     * @param consumer the exception handler
      * @param fallback the fallback route
      * @return a route for handling problems dynamically
      */
--- a/riptide-problem/src/main/java/org/zalando/riptide/problem/ProblemRoute.java
+++ b/riptide-problem/src/main/java/org/zalando/riptide/problem/ProblemRoute.java
@@ -8,8 +8,11 @@ import org.zalando.riptide.Navigators;
 import org.zalando.riptide.Route;
 
 import static org.springframework.http.MediaType.parseMediaType;
+import static org.zalando.riptide.Bindings.anyContentType;
 import static org.zalando.riptide.Bindings.on;
 import static org.zalando.riptide.Navigators.contentType;
+import static org.zalando.riptide.Route.call;
+import static org.zalando.riptide.Route.noRoute;
 import static org.zalando.riptide.RoutingTree.dispatch;
 
 public final class ProblemRoute {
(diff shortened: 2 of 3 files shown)


riptide-problem/src/main/java/org/zalando/riptide/problem/ProblemRoute.java
```diff
@@ -55,12 +58,41 @@ public static Route problemHandling() {
...
+        return problemHandling(Exceptional::propagate, fallback);
+    }
+
+    /**
+     * Produces a {@link Route route} that dispatches on the {@link Navigators#contentType() content type} and
+     * recognises {@code application/problem+json} as well as {@code application/x-problem+json} and
+     * {@code application/x.problem+json} as {@link Problem problems} and handles them given the supplied consumer.
+     * The given fallback will be used if none of the mentioned content types matches.
+     *
+     * @param fallback the fallback route
```
lukasniemeier-zalando: JavaDoc for `consumer` missing

whiskeysierra: Done

lukasniemeier-zalando:
:+1:

riptide-problem/src/main/java/org/zalando/riptide/problem/ProblemRoute.java
```diff
@@ -55,12 +58,42 @@ public static Route problemHandling() {
...
+
+        final Route route = call(Exceptional.class, consumer);
+
         return dispatch(contentType(),
                 on(PROBLEM).call(route),
                 on(X_DOT_PROBLEM).call(route),
-                on(X_DASH_PROBLEM).call(route));
+                on(X_DASH_PROBLEM).call(route),
+                anyContentType().call(fallback));
     }
```
Adeynack: 👍  that seems to be exactly the one I would need!

Adeynack: Maybe also add an example of that one in the README.

whiskeysierra: I tried to describe both individually. Finding an overloaded version that supports both should be enough for users to discover when they are discovering the API.

riptide-problem/README.md
```diff
@@ -63,6 +63,27 @@ try {
...
+
+### Fallback Route
+
+If the `ProblemRoute` fails to dispatch, e.g. because of a different media type, it will follow the default behaviour
+of Riptide and fail with a `NoRouteException` (unless a wildcard matches). This behaviour can be overridden by:
+
+```java
+http.post("/").dispatch(series(),
+    on(SUCCESSFUL).call(pass()),
+    anySeries().call(problemHandling(call(this::onUnsupportedError))));
```
whiskeysierra: Done

Adeynack:
👍

Adeynack:
Now the fun question, @whiskeysierra : when can I start using this in my prod app? aka: when will it be available on Maven?

Asking because I have a pending pull request here with a workaround. If this modification here will be deployed soon, I can change my branch and go directly with the good solution. Otherwise, I can PR my workaround and create a follow-up ticket for when this is released.

No pressure here. I really am asking just to know and plan on my side accordingly.

whiskeysierra:
@Adeynack Sorry for the late response. I released 2.4.0 that includes the changes. I hope you can still change your PR so you can make use of it.

I usually aim to get requests like this resolved in < 24 hours. Sorry that it took longer.

:heart: Thanks again for bringing up the issue!

Adeynack:
Come on @whiskeysierra :P I think that was an extremely positive response time! I signaled the issue 2 days ago and you have a full-time project at Z beside this one that keeps you busy, I'm sure ;-)

All the comm and reaction is HIGHLY APPRECIATED up here (BMO 9th floor 🙃).

tkrop:
@Adeynack wrong guess: @whiskeysierra is on vacation 😄

Adeynack:
# IT WORKS! 😍 

With this code, using the new overload:
```kotlin
            getDispatcher(http)
                .dispatch(Navigators.series(),
                    Bindings.on(HttpStatus.Series.SUCCESSFUL).call { response ->
                        deferred.resolve(Response(response, extract(response, clazz)))
                    },
                    Bindings.anySeries().dispatch(
                        Navigators.status(),
                        *errorBindings,
                        Bindings.anyStatus().call(ProblemRoute.problemHandling(
                            { deferred.reject(it as? Exception ?: it.toException()) },
                            { response, _ -> deferred.reject(RequestFailedException(response)) }))
                    )
                )
```

Everything's now green! :)

image

We have here a happy customer! 💯  😎

PR #255, opened 2017-10-19
Added happy case test for retries

### LINKED ISSUE #236: Resilience: Retries
- [x] verify that retries recover early with a successful response

whiskeysierra:
Fixes #236


lukasniemeier-zalando:
What is the race condition of https://github.com/zalando/riptide/pull/255/commits/e0e4975d3f04802eaba390097f39b8f298c29655 ? Any way to prevent that?

whiskeysierra:
Race condition in the sense that our timeouts are so small, that the calls sometimes just hit them because (my guess) that some initialization is still happening somewhere. I tried to increase them to make it more reliable.

PR #281, opened 2017-11-01
Implemented backup request resilience pattern

### LINKED ISSUE #265: Resilience: Backup Request
A new plugin that implement backup requests. Also known as *Fan Out, Quickest Reply*  and *Hedged Requests*.

whiskeysierra:
Fixes #265 

- [x] README


riptide-backup/src/main/java/org/zalando/riptide/backup/BackupRequestPlugin.java
```diff
@@ -0,0 +1,83 @@
...
+            default:
+                return execution;
+        }
+    }
+
+    private RequestExecution withBackup(final RequestExecution execution) {
+        return () -> {
+            final CompletableFuture<ClientHttpResponse> original = execution.execute();
+            final CompletableFuture<ClientHttpResponse> backup = new CompletableFuture<>();
+            original.whenComplete(cancel(delay(backup(execution, backup))));
```
jhorstmann: Should the original be canceled in case the backup completes first?

whiskeysierra: I'm not cancelling the backup request directly, but rather the schedule to send it. Cancelling an ongoing request will have virtually no impact on the server side, since it's unable to detect it until the response will be written back.

riptide-backup/src/main/java/org/zalando/riptide/backup/BackupRequestPlugin.java
```diff
@@ -0,0 +1,83 @@
...
+            }
+        };
+    }
+
+    private ScheduledFuture<?> delay(final Runnable runnable) {
+        return scheduler.schedule(runnable, delay, unit);
+    }
+
+    private static <T> BiConsumer<T, Throwable> cancel(final Future<?> future) {
+        return (result, throwable) -> future.cancel(false);
```
jhorstmann: For network requests cancelling with interruption might make sense

riptide-backup/src/main/java/org/zalando/riptide/backup/BackupRequestPlugin.java
whiskeysierra: Also, cancelling a `CompletableFuture` is basically useless, since it has no connection to the thread doing the actual work and therefore there is nothing to cancel.

riptide-backup/src/main/java/org/zalando/riptide/backup/BackupRequestPlugin.java
whiskeysierra: This is cancelling the scheduled future that starts the request, not the request itself.

riptide-backup/src/main/java/org/zalando/riptide/backup/BackupRequestPlugin.java
jhorstmann: ok, got it. The actual request is async again and so currently won't be cancelled.

One more suggestion, introduce a variable for the result of `delay` to make it clearer it's not a pure function but schedules a thread as a side-effect. Suggested name `scheduledBackup`?

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/RiptideRegistrar.java
```diff
@@ -289,6 +279,47 @@ private String findFaultClassifier(final String id) {
...
+    private BeanMetadataElement registerScheduler(final String id, final Client client) {
+        // we allow users to use their own ScheduledExecutorService, but they don't have to configure tracing
+        return trace(registry.registerIfAbsent(id, ScheduledExecutorService.class,
+                () -> {
+                    final CustomizableThreadFactory threadFactory = new CustomizableThreadFactory(
+                            "http-" + id + "-scheduler-");
+                    threadFactory.setDaemon(true);
+
+                    return genericBeanDefinition(ScheduledThreadPoolExecutor.class)
+                            .addConstructorArgValue(client.getMaxConnectionsTotal() / 5)
```
lukasniemeier-zalando: Single thread?

lukasniemeier-zalando:
:+1:

PR #286, opened 2017-11-10
Attempt to allow futures to be cancellable

### LINKED ISSUE #282: Cancelability
http://www.nurkiewicz.com/2015/03/completablefuture-cant-be-interrupted.html
http://www.nurkiewicz.com/2013/05/java-8-completablefuture-in-action.html
https://github.com/vsilaev/tascalate-concurrent
https://github.com/lukas-krecan/completion-stage

- [x] Support Java 9 in a backwards compatible way, i.e. `newIncompleteFuture`
- [ ] Test cancellation of backup requests
- [ ] Test cancellation of timeout requests
- [ ] Test cancellation of retried requests
- [ ] Test cancellation of exceptional futures

whiskeysierra:
Fixes #282

I'm not really happy with this, but I can't find a better solution as long as we're on Java 8.


riptide-core/src/main/java/org/zalando/riptide/AbstractCancelableCompletableFuture.java
```diff
@@ -0,0 +1,22 @@
+package org.zalando.riptide;
+
+import java.util.concurrent.CompletableFuture;
+
+import static org.zalando.riptide.CancelableCompletableFuture.preserveCancelability;
+
+public abstract class AbstractCancelableCompletableFuture<T> extends CompletableFuture<T> {
```
tkrop: I this this is one level more abstract than necessary. Especially, its only method strongly couples it to its default implementation.

tkrop: Oh, I see it is because of Java 9.

riptide-core/src/main/java/org/zalando/riptide/Requester.java
```diff
@@ -109,7 +113,18 @@ public final Requester contentType(final MediaType contentType) {
 
                 // TODO why not return CompletableFuture<ClientHttpResponse> here?
                 // we need a CompletableFuture<Void>
-                return future.thenApply(response -> null);
+
+                // TODO: replace with thenApply call in Java 9
+                final CompletableFuture<Void> result = preserveCancelability(future);
+                future.whenComplete((response, throwable) -> {
```
tkrop: `future.whenComplete(forwardTo(result))`?

tkrop:
@whiskeysierra I must admit I still do not get the problem exactly. Up to here is this only for Java 9 adaptation?

I do not understand how this patch finally helps to solve the problem of cancel. It should be clear, that a result supplier has to listen to `future.isCancelled()`, to know that it is not supposed to work on this task any longer. The same holds for the subsequent tasks submitted to `future.thanApply()` etc.

In general, I expect `CompleteableFuture` to propagate the cancellation back to the source task to allow for correct cancellation handling. As far as I can see this is done by `CompleteableFuture` for all completion stages in a way that allows to cancel the pipeline between task calls. So everything works fine as long as tasks are progressing. So what is left for cancellation are in my opinion the following situations:

1. The task is scheduled in a queue of an executor services,
2. The task is running within a thread (possibly blocked).

While for (1) the task can be removed from the queue or stopped before execution, (2) can only be handled by interrupting the execution task. As far as I can see no executor service implementation offers the full ability of cancellation - even the `ForkJoinTask` explicitly state that interrupts are not used for cancellation.

Thus, I do not see any other way to solve this problem as to make the tasks aware of their future - currently.

whiskeysierra:
> I do not understand how this patch finally helps to solve the problem of cancel. It should be clear, that a result supplier has to listen to future.isCancelled(), to know that it is not supposed to work on this task any longer.

That method returns a boolean, hence you can't really *listen* to it. You can try to inspect it at certain times, but it's not a trigger that you can attach a callback to or anything.

> The same holds for the subsequent tasks submitted to future.thanApply() etc.

Not really. You can complete a future at most once and any callbacks that are registered for the happy-path (`thenApply`, `thenRun`, `thenAccept`, ...) will **not** be triggered in the first place.

> In general, I expect CompleteableFuture to propagate the cancellation back to the source task to allow for correct cancellation handling.

This is not the case and it's exactly the problem that I'm (partially) trying to address in this PR. Please read the links mentioned in the corresponding issue. TL;DR: `CompletableFuture` is **not** bound to a specific thread. Any number of threads may or may not try to complete it. See also:

> Limited cancel() in case of CompletableFuture is not a bug, but a design decision. CompletableFuture is not inherently bound to any thread, while Future almost always represents background task. It's perfectly fine to create CompletableFuture from scratch (new CompletableFuture<>()) where there is simply no underlying thread to cancel.
>
> http://www.nurkiewicz.com/2015/03/completablefuture-cant-be-interrupted.html

> As far as I can see no executor service implementation offers the full ability of cancellation

What makes you believe that? `ThreadPoolExecutor` should most certainly support interruptions. It's up to the code that performs the blocking operation to correctly respond to an interrupt.

tkrop:
> That method returns a boolean, hence you can't really *listen* to it.

You are precise as ever :). Yes, any task has to poll this information.

> This is not the case and it's exactly the problem that I'm (partially) trying to address in this PR. ...

WTF, I see this bug :open_mouth:.

>>  As far as I can see no executor service implementation offers the full ability of cancellation
>
> What makes you believe that?

Because `Runnable` and `Callable` as common task input for executor service have no `cancel`, and `ForkJoinTask`, which has a `cancel`, explicitly states, that it will not interrupt on `cancel`. So if the task is running it is still responsibility of the task programmer to check for this condition.

whiskeysierra:
> Because Runnable and Callable as common task input for executor service have no cancel

They don't need to. As long as they are blocking, they can catch InterruptedException and deal with it.

riptide-core/src/main/java/org/zalando/riptide/CancelableCompletableFuture.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    public boolean cancel(final boolean mayInterruptIfRunning) {
+        cause.cancel(mayInterruptIfRunning);
+        super.cancel(mayInterruptIfRunning);
+        return isCancelled();
+    }
+
+    public static <T> CompletableFuture<T> preserveCancelability(final Future<?> future) {
+        return new CancelableCompletableFuture<>(future);
```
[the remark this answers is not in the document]
whiskeysierra: The second one is a Spring API, and I don't want to make this class depend on spring.

riptide-core/src/main/java/org/zalando/riptide/Requester.java
whiskeysierra: Won't work, since this is a poor man's `thenApply`, i.e. I'm passing `null` instead of response.

riptide-core/src/main/java/org/zalando/riptide/Requester.java
```diff
@@ -118,7 +133,11 @@ public final Requester contentType(final MediaType contentType) {
         private CompletableFuture<ClientHttpResponse> send() throws IOException {
             final AsyncClientHttpRequest request = createRequest();
             worker.write(request, entity);
-            return new ListenableToCompletableFutureAdapter<>(request.executeAsync());
+            final ListenableFuture<ClientHttpResponse> original = request.executeAsync();
+
+            final CompletableFuture<ClientHttpResponse> future = preserveCancelability(original);
+            original.addCallback(future::complete, future::completeExceptionally);
```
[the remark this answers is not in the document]
whiskeysierra: It's similar, but `original` is a `ListenableFuture` and it doesn't support `whenComplete`.

riptide-core/src/main/java/org/zalando/riptide/Requester.java
tkrop: True here. But why?

riptide-core/src/main/java/org/zalando/riptide/Requester.java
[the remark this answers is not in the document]
tkrop: To bad, I see a great potential of simplification by using `whenComplete(forwardTo())`. At least for me it is much easier to understand how forwarding of results works together with cancelling. Especially, if it would be integrated into `preserveCancelabiliy()`, this would maker absolutely clear how the forwarding works and ensure that it is done properly.

riptide-core/src/main/java/org/zalando/riptide/Requester.java
whiskeysierra: Because the API expects a `CompletableFuture<Void>`. I already have the plan to change this in version 3.0, see #290.

riptide-core/src/main/java/org/zalando/riptide/Requester.java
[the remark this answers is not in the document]
whiskeysierra: I have usages in the code bases that only need `preserveCancelability` as well as ones that only use `forwardTo`. I hope this will become a lot simpler if we can leverage Java 9 to its full extent.

tkrop:
Okay than. Do you want to merge?

tkrop:
👍

PR #297, opened 2017-11-22
Introduced Types factory

### LINKED ISSUE #294: Central type token factory
`Route` contains static factory methods to create type tokens for `List<T>` and `ResponseEntity<T>`. Those should be moved into a special class, e.g. `Typing`.

whiskeysierra:
Fixes #294


lukasniemeier-zalando:
:+1:

PR #298, opened 2017-11-22
Denied second captures

### LINKED ISSUE #292: Capture should deny second capture
Could throw an `IllegalStateException`

whiskeysierra:
Fixes #292


riptide-capture/src/main/java/org/zalando/riptide/capture/DefaultCapture.java
```diff
@@ -19,8 +19,12 @@
 
     @Override
     public void tryAccept(@Nullable final T input) {
-        // TODO fail on second capture?
-        reference.compareAndSet(null, Optional.ofNullable(input));
+        final boolean captured = reference.compareAndSet(null, Optional.ofNullable(input));
+        final boolean alreadyCaptured = !captured;
```
AlexanderYastrebov: Not sure if this one improves readability, I would put negation into if condition

whiskeysierra: Done

lukasniemeier-zalando:
:+1:

PR #299, opened 2017-11-22
Introduced specialized capture exception

### LINKED ISSUE #293: Capture should throw specific/custom exception type
Right now a `NoSuchElementException` is thrown when no value was captured. Changing this to a custom exception type (that extends `NoSuchElementException`) could improve the usability/readability.

whiskeysierra:
Fixes #293


lukasniemeier-zalando:
:+1:

PR #301, opened 2017-11-24
Riptide Client Testing

lukasniemeier-zalando:
Introducing `@RiptideClientTest` - a new spring-boot meta-annotation to support testing riptide-spring-boot-starter generated Riptide `Http` clients.

Currently testing is a bit [tedious and repetitive](https://github.com/zalando/riptide#testing). It can be shorter:

```
@Component
public class GatewayService {

    @Autowired
    @Qualifier("example")
    private Http http;

    void remoteCall() {
        http.get("/bar").dispatch(status(), on(OK).call(pass())).join();
    }
}

@RunWith(SpringRunner.class)
@RiptideClientTest(GatewayService.class)
public class RiptideTest {

    @Autowired
    private GatewayService client;

    @Autowired
    private MockRestServiceServer server;

    @Test
    public void shouldAutowireMockedHttp() throws Exception {
        server.expect(requestTo("https://example.com/bar")).andRespond(withSuccess());
        client.remoteCall()
        server.verify();
    }
}
```

`@RiptideClientTest` is based on [`@RestClientTest`](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/test/autoconfigure/web/client/RestClientTest.html).

Currently `@RiptideClientTest` and its facilities are tied to riptide-spring-boot-starter. In order to allow usage with riptide-core only, it may make sense to extract the components to a dedicated artifact. Instead of modifying the `RiptideRegistar` directly, one would need to make sure that properly named `AsyncClientHttpRequestFactory` are registered beforehand (expose `Registry#generateBeanName`?). Added benefit is that we wouldn't need any `provided` testing dependencies anymore. On the other hand I am not sure whether anybody wants to / needs to use this without the starter...


riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/testing/RiptideMockRestServiceServerAutoConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+    @Bean
+    public RequestExpectationManager requestExpectationManager() {
+        return new SimpleRequestExpectationManager();
+    }
+
+    @Bean
+    public MockRestServiceServer mockRestServiceServer(final RequestExpectationManager manager) throws Exception {
+        final Constructor<MockRestServiceServer> ctr = MockRestServiceServer.class.getDeclaredConstructor(RequestExpectationManager.class);
+        ctr.setAccessible(true);
+        return ctr.newInstance(manager);
```
lukasniemeier-zalando: Hello Darkness My Old Friend

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/RiptideSettings.java
```diff
@@ -23,6 +23,11 @@
     private Defaults defaults = new Defaults();
     private GlobalOAuth oauth = new GlobalOAuth();
     private Map<String, Client> clients = new LinkedHashMap<>();
+    private Boolean mocked = false;
+
+    public void setMocked(final Boolean mocked) {
```
whiskeysierra: You should get this for free.

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/RiptideSettings.java
```diff
@@ -23,6 +23,11 @@
     private Defaults defaults = new Defaults();
     private GlobalOAuth oauth = new GlobalOAuth();
     private Map<String, Client> clients = new LinkedHashMap<>();
+    private Boolean mocked = false;
```
whiskeysierra: Why not `boolean`?

whiskeysierra: Putting this here technically allows users to do this:

```yaml
riptide:
  mocked: true
```

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/testing/RiptideClientTest.java
```diff
@@ -0,0 +1,41 @@
...
+})
+public @interface RiptideClientTest {
+
+    /**
+     * Specifies the components to test. May be left blank if components will be manually
+     * imported or created directly.
+     *
+     * @return the components to test
+     */
+    @AliasFor(annotation = RestClientTest.class, attribute = "value")
```
whiskeysierra: Any other attributes that are worth having as an alias?

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/testing/RiptideClientTest.java
```diff
@@ -0,0 +1,41 @@
...
+import java.lang.annotation.Target;
+
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Inherited
+@RestClientTest
+@AutoConfigureMockRestServiceServer(enabled = false) // see RiptideClientAutoConfiguration
+@TestPropertySource(properties = {
+        "riptide.mocked: true"
```
whiskeysierra: There it is.

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/testing/ExpectingHttpRequestFactory.java
```diff
@@ -0,0 +1,46 @@
...
+import org.springframework.http.client.ClientHttpRequest;
+import org.springframework.http.client.ClientHttpRequestFactory;
+import org.springframework.http.client.ClientHttpResponse;
+import org.springframework.mock.http.client.MockAsyncClientHttpRequest;
+import org.springframework.test.web.client.RequestExpectationManager;
+
+import java.io.IOException;
+import java.net.URI;
+
+public class ExpectingHttpRequestFactory implements AsyncClientHttpRequestFactory, ClientHttpRequestFactory {
```
whiskeysierra: Can't we use the existing implementation?

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/RiptideSettings.java
lukasniemeier-zalando: Top Level of the Settings currently only has setters generated.

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/RiptideSettings.java
lukasniemeier-zalando: We can unbox - no particular reason to use Boolean

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/testing/RiptideClientTest.java
lukasniemeier-zalando: Also referencing your comment above: any other method to inject the "mock mode" is obviously fine - this has been just the fastest way :grin:

whiskeysierra:
Summary of this PR contains `RiptideGatewayTest` rather than `RiptideClientTest`.

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/testing/ExpectingHttpRequestFactory.java
lukasniemeier-zalando: I think it is private. Yeah I know I am already using the private constructor of the server, but this guy here is 10 straight forward lines of code

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/RiptideSettings.java
whiskeysierra: You mean getters, right? Isn't the constructor enough?

whiskeysierra:
> Instead of modifying the RiptideRegistar directly, one would need to make sure that properly named AsyncClientHttpRequestFactory are registered beforehand (expose Registry#generateBeanName?). Added benefit is that we wouldn't need any provided testing dependencies anymore. On the other hand I am not sure whether anybody wants to / needs to use this without the starter...

I think that would be cleaner, even without the dedicated artifact. Leaving the Registrar purely for production would be easier to understand for future changes.

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/RiptideSettings.java
lukasniemeier-zalando: I did not get it working without this `setter` somehow. He does not find a proper "writable" property path.

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/testing/RiptideClientTest.java
lukasniemeier-zalando: I haven't used any other on `@RestClientTest`, but we may decided to just forward all of it?

jhorstmann:
Are OAuth access tokens also mocked? Would be nice to avoid any external calls in the tests.

whiskeysierra:
> Are OAuth access tokens also mocked? Would be nice to avoid any external calls in the tests.

Not explicitly. The `AccessTokens` instance will try to generate tokens, but they will never be used, since the interceptor that would inject them is not being used (it's for the real Apache HTTP client).

whiskeysierra:
The test setup could just register a `AccessTokens` bean that does nothing.

lukasniemeier-zalando:
I've just pushed an alternative implementation that relies on the original trick from the `README` and does not use a property to modify the behavior of the `RiptideRegistar`.

The example above changes slightly, as there have to be one mocked server per client:

```
@Component
public class GatewayService {

    @Autowired
    @Qualifier("example")
    private Http http;

    void remoteCall() {
        http.get("/bar").dispatch(status(), on(OK).call(pass())).join();
    }
}

@RunWith(SpringRunner.class)
@RiptideClientTest(GatewayService.class)
public class RiptideTest {

    @Autowired
    private GatewayService client;

    @Autowired
    @Qualifier("example") // this is now needed
    private MockRestServiceServer server;

    @Test
    public void shouldAutowireMockedHttp() throws Exception {
        server.expect(requestTo("https://example.com/bar")).andRespond(withSuccess());
        client.remoteCall()
        server.verify();
    }
}
```

One important (?) thing to note is that the `RestTemplate` won't be redirected to the mock request factory. In order to do so we'd have to register another dedicated `MockRestServiceServer` - I guess this is not worth it as then we'd have multiple `MockRestServiceServer` beans with the same client qualifier.

lukasniemeier-zalando:
> The test setup could just register a `AccessTokens` bean that does nothing.

:+1:

lukasniemeier-zalando:
> > The test setup could just register a `AccessTokens` bean that does nothing.

> :+1: 

Actually this is not needed. The `AccessTokens` integration works via interceptors of the `HttpClient` of the request factory. And the latter we mock.

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/TestRiptideRegistrar.java
```diff
@@ -0,0 +1,55 @@
...
+            log.debug("Client [{}]: Registering mocked AsyncClientHttpRequestFactory", id);
+            final BeanDefinitionBuilder factory = genericBeanDefinition(AsyncClientHttpRequestFactory.class);
+            factory.addDependsOn(serverId);
+            factory.setFactoryMethodOnBean("getAsyncRequestFactory", templateId);
+            return factory;
+        });
+    }
+
+    private String registerAsyncRestTemplate(final String id) {
+        return registry.registerIfAbsent(id, AsyncRestTemplate.class, () -> {
```
whiskeysierra: That will incorrectly replace the `AsyncRestTemplate` that we would register in the `DefaultRegistrar`. It would miss plugins as well as the base url.

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/TestRiptideRegistrar.java
```diff
@@ -0,0 +1,55 @@
...
+            log.debug("Client [{}]: Registering AsyncRestTemplate", id);
+            return genericBeanDefinition(AsyncRestTemplate.class);
+        });
+    }
+
+    private String registerMockRestServiceServer(final String id, final String templateId) {
```
whiskeysierra: Why don't we register this once as a singleton?

whiskeysierra: ... and share it across clients?

lukasniemeier-zalando: Hm I remember that we talked about having this as a singleton vs. per-client and somehow I was under the impression that I gave in to have it per-client :grimacing: I believe the singleton route is more intuitive either way!

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/TestRiptideRegistrar.java
lukasniemeier-zalando: Fixed

riptide-spring-boot-starter/src/test/java/org/zalando/riptide/spring/testing/TestService.java
```diff
@@ -0,0 +1,30 @@
...
+@Component
+public class TestService {
+
+    private final Http http;
+    private final RestTemplate restTemplate;
+
+    @Autowired
+    public TestService(@Qualifier("example") final Http http, @Qualifier("example") final RestTemplate restTemplate) {
+        this.http = http;
+        this.restTemplate = restTemplate;
```
whiskeysierra: Not used?

whiskeysierra:
> Actually this is not needed. The AccessTokens integration works via interceptors of the HttpClient of the request factory. And the latter we mock.

Access tokens are never used, but the default implementation will spawn a thread and poll tokens in the background. That could be disabled.

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/RiptideClientTest.java
```diff
@@ -0,0 +1,36 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Inherited
+@RestClientTest
+@AutoConfigureMockRestServiceServer(enabled = false) // will be registered per client
```
whiskeysierra: Comment is out-dated. Can you describe here why we need this?

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/TestRiptideRegistrar.java
```diff
@@ -0,0 +1,53 @@
...
+    private final RiptideSettings settings;
+
+    @Override
+    public void register() {
+        final String templateId = registerMockAsyncRestTemplate();
+        final String serverId = registerMockRestServiceServer(templateId);
+        settings.getClients().forEach((id, client) -> registerAsyncClientHttpRequestFactory(id, templateId, serverId));
+    }
+
+    private String registerMockAsyncRestTemplate() {
```
whiskeysierra: Couldn't you register the template and the server using traditional `@Bean` methods in the `RiptideTestAutoConfiguration`?

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/RiptideClientTest.java
lukasniemeier-zalando: Ups will do

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/TestRiptideRegistrar.java
lukasniemeier-zalando: Technically yes, but I haven't found a cool way to still reference these beans in here on registering the request factories. Just by some explicit bean name? Any idea? I feel those `2 + (1 * n)` beans should go together.

riptide-spring-boot-starter/src/test/java/org/zalando/riptide/spring/testing/TestService.java
lukasniemeier-zalando: Fixed

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/TestRiptideRegistrar.java
whiskeysierra: Explicit bean name should work. I'd argue that the configuration binds those 2+n beans together. The Registrar is just our way to dynamically register the n beans.

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/RiptideTestAutoConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+    @Configuration
+    static class MockConfiguration {
+
+        @Bean(name = TEMPLATE_BEAN_NAME)
+        AsyncRestTemplate mockAsyncRestTemplate() {
+            return new AsyncRestTemplate();
+        }
+
+        @Bean(name = SERVER_BEAN_NAME)
+        MockRestServiceServer mockRestServiceServer(@Qualifier("_mockAsyncRestTemplate") final AsyncRestTemplate template) {
```
whiskeysierra: The qualifier could use the `TEMPLATE_BEAN_NAME`.

lukasniemeier-zalando: Ups forgot to replace

whiskeysierra:
:+1:

lukasniemeier-zalando:
In a follow-up I will add a paragraph to the `README`s

RohanNg:
Hi,

We had a use case where in `remoteCall` multiple requests are dispatched to a http endpoint. We have a problem writing test because the ordering of the http call cannot be relaxed, while it is desirable to ignore the ordering. 

This could be probable be done by `MockRestServiceServer.bindTo(restTemplate).ignoreExpectOrder(true).build()`. However we don't know how to get the produced server instance into use. 

Could you provide some hint how to do that?

whiskeysierra:
No easy way to do that right now. Options that I see:

1. Use reflection to manipulate the `expectationManager` and replace it with an `UnorderedRequestExpectationManager`
2. Open a pull request and introduce an option (property?) to change the behavior of this line:
https://github.com/zalando/riptide/blob/0f53529a69c864203ced5fa067dcf3d42dba0e26/riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/RiptideTestAutoConfiguration.java#L39
3. Change the line mentioned above, introduce a `@ConditionalOnBean(MockRestServiceServer.class)` and then override it in your auto configuration. Properly the cleanest solution: :heavy_check_mark:

Issue #312, opened 2017-12-18
Failsafe Starter: Endless retries by default


lukasniemeier-zalando:
Combining the starter with `riptide-failsafe` results in endless retries by default.

```
    my-service:
      base-url: https://foo.bar
```

In above's example user has not configured any `retry` configuration. Therefore default `retryPolicy` of `RetryPolicyFactoryBean` is used, which is configured to allow endless retries without delay (`maxRetries = -1`).

This seems unexpected, I'd expect **no delay** to be configured (`maxRetries = 0`). As soon as any other field is specified (e.g. `maxDuration`) delays should be allowed.

What do you think?

whiskeysierra:
That was my intent.

tkrop:
@whiskeysierra hmm, in my opinion an infinite retry is not a very sensible default. As an effect failures will result in resource consumption instead of backing off. A better way is to end the requests after a low number of retries with some default random backoff period. In cases where I used this concept the first retry could happen after even some milliseconds, while the second retry happened after a view seconds or a minute - to cover for handover and traffic switching situations.

Any other behavior requires manual configuration that takes the special knowledge about the system failures and their behaviors into account.

whiskeysierra:
Sorry, what I meant is that it should not retry by default. See Lukas'
suggestion above

PR #315, opened 2018-01-05
GH-312 No retries if not configured

lukasniemeier-zalando:
Fixes GH-312.


whiskeysierra:
I'll try to summarize to see whether I understood it correctly:

1. If the `retry` key is missing, we don't retry (max retries = 0)
2. Else
    1. If `max-retries` is present, use it (max retries = anything; -1, 0 or else)
    2. Else, retry forever (max retries = -1)

lukasniemeier-zalando:
Yup matches the three test cases in https://github.com/zalando/riptide/pull/315/files#diff-90736be94b1fc612595e8ffe7d878b90

whiskeysierra:
:+1:

PR #316, opened 2018-01-05
Added support to handover work to original thread pool

### LINKED ISSUE #283: Schedulers should do minimal work
Follow-up from #279 

1. Executing a request should be as short, i.e. as async, as possible
2. Callbacks should not be executed by the scheduler. That may happen if the scheduler thread completes the future.

whiskeysierra:
Fixes #283


lukasniemeier-zalando:
:+1:

PR #320, opened 2018-01-10
Registered a custom task scheduler, if none was present

### LINKED ISSUE #319: Request hangs when retries or circuit breakers are configured
## Behaviour

Multiple users reported the following behaviour:

> The application hangs forever when making a remote request.

All users had two things in common. They had retries and/or a circuit breaker configured:
```yaml
riptide:
  clients:
    example:
      base-url: http://httpbin.org/
      retry:
        fixed-delay: 50 milliseconds
      circuit-breaker:
        failure-threshold: 3 out of 5
```

They also made use of `@EnableScheduling` and `@Scheduled` annotation.

## Cause

Retries and circuit breaker require a scheduler (`ScheduledThreadPoolExecutor`), so the `riptide-spring-boot-starter` registers one per client (1 in this case). But then Spring does the following:

> By default, will be searching for an associated scheduler definition: either a unique TaskScheduler bean in the context, or a TaskScheduler bean named "taskScheduler" otherwise; the same lookup will also be performed for a ScheduledExecutorService bean. If neither of the two is resolvable, a local single-threaded default scheduler will be created and used within the registrar.
>
> https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/scheduling/annotation/EnableScheduling.html

In other words, Spring will silently use our scheduler for `@Scheduled` methods and since our scheduler is single threaded, there is no thread left for the retries/circuit breaker.

## Possible workarounds

1. Register a custom scheduler called `taskScheduler`:
```java
@Bean(destroyMethod = "shutdown")
public ScheduledExecutorService taskScheduler() {
    return Executors.newScheduledThreadPool(Runtime.getRuntime().availableProcessors());
}
```
2. Use the [Tracer Spring Boot Starter](https://github.com/zalando/tracer#spring-boot-starter), as it will register such a scheduler for you (+ it will enable call tracing):
```xml
<dependency>
	<groupId>org.zalando</groupId>
	<artifactId>tracer-spring-boot-starter</artifactId>
	<version>0.11.2</version>
</dependency>
```
3. Define more than one client in the `application.yaml` so that no unique scheduler exists in the context, which in turn forces Spring to create a new scheduler ([be aware, that one also has a size of 1](https://stackoverflow.com/questions/29796651/what-is-the-default-scheduler-pool-size-in-spring-boot))
```yaml
foo:
  circuit-breaker:
    failure-threshold: 1 out of 10
```

whiskeysierra:
Fixes #319 

Based on the idea from @lukasniemeier-zalando. We register a task scheduler (size = # of CPUs), if and only if no scheduler was registered before. We let `TracerAutoConfiguration` take precedence.

### DIFF
--- a/riptide-backup/src/test/java/org/zalando/riptide/backup/BackupRequestPluginTest.java
+++ b/riptide-backup/src/test/java/org/zalando/riptide/backup/BackupRequestPluginTest.java
@@ -73,7 +73,7 @@ public final class BackupRequestPluginTest {
 
         unit.get("/bar")
                 .call(pass())
-                .get(2500, TimeUnit.MILLISECONDS);
+                .get(3, SECONDS);
     }
 
     @Test(expected = IllegalStateException.class)
--- a/riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/DefaultRiptideRegistrar.java
+++ b/riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/DefaultRiptideRegistrar.java
@@ -218,15 +218,11 @@ final class DefaultRiptideRegistrar implements RiptideRegistrar {
         }
 
         if (client.getRetry() != null || client.getCircuitBreaker() != null) {
-            plugins.add(ref(registry.registerIfAbsent(id, FailsafePlugin.class, () -> {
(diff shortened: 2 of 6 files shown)


lukasniemeier-zalando:
:+1:

PR #321, opened 2018-01-12
Changed classifier to inspect whole chain of throwables and not just the root cause

### LINKED ISSUE #310: DefaultFaultClassifier classifies root cause but not throwable itself
From the code it looks like that `DefaultFaultClassifier` classifies root cause but not throwable itself.

whiskeysierra:
Fixes #310

### DIFF
--- a/riptide-faults/src/main/java/org/zalando/riptide/faults/DefaultFaultClassifier.java
+++ b/riptide-faults/src/main/java/org/zalando/riptide/faults/DefaultFaultClassifier.java
@@ -1,8 +1,8 @@
 package org.zalando.riptide.faults;
 
-import com.google.common.base.Throwables;
-
+import java.util.Objects;
 import java.util.function.Predicate;
+import java.util.stream.Stream;
 
 final class DefaultFaultClassifier implements FaultClassifier {
 
@@ -14,13 +14,12 @@ final class DefaultFaultClassifier implements FaultClassifier {
 
     @Override
     public Throwable classify(final Throwable throwable) {
-        final Throwable root = Throwables.getRootCause(throwable);
-
-        if (isTransient.test(root)) {
-            return new TransientFaultException(throwable);
-        }
-
-        return throwable;
+        return Stream.iterate(throwable, Throwable::getCause)
+                .filter(Objects::nonNull)
+                .findFirst()
+                .filter(isTransient)
+                .<Throwable>map(TransientFaultException::new)
+                .orElse(throwable);
     }
 
 }
--- a/riptide-faults/src/main/java/org/zalando/riptide/faults/DefaultFaultClassifier.java
+++ b/riptide-faults/src/main/java/org/zalando/riptide/faults/DefaultFaultClassifier.java
@@ -1,8 +1,8 @@
 package org.zalando.riptide.faults;
 
-import java.util.Objects;
+import com.google.common.base.Throwables;
+
 import java.util.function.Predicate;
-import java.util.stream.Stream;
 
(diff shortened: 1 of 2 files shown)


riptide-faults/src/main/java/org/zalando/riptide/faults/DefaultFaultClassifier.java
```diff
@@ -14,13 +14,12 @@ public DefaultFaultClassifier(final Predicate<Throwable> isTransient) {
...
     @Override
     public Throwable classify(final Throwable throwable) {
-        final Throwable root = Throwables.getRootCause(throwable);
-
-        if (isTransient.test(root)) {
-            return new TransientFaultException(throwable);
-        }
-
-        return throwable;
+        return Stream.iterate(throwable, Throwable::getCause)
```
AlexanderYastrebov: Cool! I did not know about iterate

riptide-faults/src/main/java/org/zalando/riptide/faults/DefaultFaultClassifier.java
```diff
@@ -14,13 +14,12 @@ public DefaultFaultClassifier(final Predicate<Throwable> isTransient) {
...
+                .filter(Objects::nonNull)
+                .findFirst()
```
AlexanderYastrebov: Looks like `.findFirst()` and `.filter(isTransient)` should be swapped

whiskeysierra: That actually will break if a non-transient throwable is given without any cause.

riptide-faults/src/test/java/org/zalando/riptide/faults/FaultClassifierTest.java
```diff
@@ -35,6 +35,13 @@ public void shouldNotClassifyGenericSSLHandShakeException() {
         assertNotTransient(new SSLHandshakeException("No hands, no cookies"));
     }
 
+    @Test
+    public void shouldClassifyAsTransientWithNonTransientRootCause() {
```
AlexanderYastrebov: I think we need two additional tests, beside testing root cause match:
1. test that verifies matching throwable itself
2. test that verifies intermediate cause (`new Exception("outer", new Exception("intermediate", new Exception("inner")))`)

whiskeysierra: > test that verifies matching throwable itself

See line 20.

riptide-faults/src/main/java/org/zalando/riptide/faults/DefaultFaultClassifier.java
AlexanderYastrebov: Why? iterate return throwable itself as first element

whiskeysierra: Yes, but if it's not transient then the stream is empty and findFirst will throw.

AlexanderYastrebov: > Returns an Optional describing the first element of this stream, or an empty Optional if the stream is empty.

whiskeysierra: I think I misinterpreted an exception that I saw during tests.

AlexanderYastrebov:
:+1:

PR #322, opened 2018-01-15
Prevent overwriting TaskScheduler of Tracer

lukasniemeier-zalando:
Follow up to https://github.com/zalando/riptide/pull/320 - ensures that Tracer's `AutoConfiguration` runs before Riptide's.

Feel free to take this as an inspiration only in case you have more ideas on how to simplify the `AutoConfiguration` classes.


lukasniemeier-zalando:
Apparently https://github.com/zalando/tracer/pull/71 also fixes the same problem.

riptide-spring-boot-starter/src/main/java/org/zalando/riptide/spring/RiptideAutoConfiguration.java
```diff
@@ -24,6 +24,7 @@
         "org.springframework.boot.autoconfigure.jackson.JacksonAutoConfiguration",
         "org.zalando.logbook.spring.LogbookAutoConfiguration",
         "org.zalando.tracer.spring.TracerAutoConfiguration",
+        "org.zalando.tracer.spring.TracerSchedulingAutoConfiguration",
 })
 public class RiptideAutoConfiguration {
```
whiskeysierra: To clean this up, we should remove the `AutoConfigureAfter` from the inner configuration and move the `AutoConfigureBefore` here as well.

whiskeysierra:
:+1:

PR #325, opened 2018-01-17
Introduces staged builder and removed configurer

### LINKED ISSUE #295: Change HttpBuilder to a staged builder
Request factory is in fact a mandatory dependency, but the API allows to create an `Http` instance without it, but fails at runtime. Changing the API would make this a compile time error.

See https://immutables.github.io/immutable.html#staged-builder. Whether we should use `Immutables` for this, that's up to discussion.

whiskeysierra:
Fixes #295


riptide-core/src/main/java/org/zalando/riptide/Http.java
```diff
@@ -56,8 +63,34 @@
...
     Requester execute(HttpMethod method);
 
-    static HttpBuilder builder() {
+    static RequestFactoryStage builder() {
         return new DefaultHttpBuilder();
     }
 
+    interface RequestFactoryStage {
+        ConfigurationStage requestFactory(AsyncClientHttpRequestFactory requestFactory);
+        default ConfigurationStage simpleRequestFactory(final ExecutorService executor) {
```
lukasniemeier-zalando: What about exposing an overload without any parameter i.e. `simpleRequestFactory()` ?

whiskeysierra: I don't want to manage an executor service myself. Clients should take care of this, including stopping/shutting down.

whiskeysierra: I'm actually thinking about moving the whole method to a static factory and out of the builder interface. But I have a couple of other PRs lined up and this will have to wait.

PR #404, opened 2018-06-21
Refactored metrics plugin and modified docs

### LINKED ISSUE #403: Metrics readme incorrectly claims that micrometer integrates with Spring Boot 1.x metrics
## Expected Behavior

Readme shouldn't claim that `/metrics` works for Spring Boot 1.x.

## Actual Behavior

Readme is incorrect.

whiskeysierra:
## Motivation and Context

Fixes #403

## Types of changes

## Checklist:


riptide-metrics/src/main/java/org/zalando/riptide/metrics/MetricsPlugin.java
```diff
@@ -10,33 +12,63 @@
...
-                        if (nonNull(response)) {
-                            measurement.record(response);
-                        }
-                    }));
+            final CompletableFuture<ClientHttpResponse> future;
+
+            try {
+                future = execution.execute();
+            } catch (final IOException e) {
+                measurement.record(null, e);
```
whiskeysierra: How do I test this?

riptide-metrics/src/main/java/org/zalando/riptide/metrics/MetricsPlugin.java
```diff
@@ -48,23 +80,54 @@ public RequestExecution prepare(final RequestArguments arguments, final RequestE
...
+
+    private String status(@Nullable final ClientHttpResponse response) {
+        if (response == null) {
+            return "CLIENT_ERROR";
         }
 
+        try {
+            return String.valueOf(response.getRawStatusCode());
+        } catch (final IOException e) {
+            return "IO_ERROR";
```
whiskeysierra: Any idea how to cover/enforce this?

lukasniemeier-zalando: Or we just let the exception propagate and don't measure in these cases

riptide-metrics/src/main/java/org/zalando/riptide/metrics/MetricsPlugin.java
lukasniemeier-zalando: super specific unit test with super specific mocks... :deciduous_tree:

whiskeysierra:
I decided to get rid of the timing of failed requests, since I don't really see a big use case right now. We may need to revisit that decision.

lukasniemeier-zalando:
:+1:

PR #406, opened 2018-06-25
Fixed single-digit issue

whiskeysierra:
## Description

Fixes an issue with the `RetryAfterDelayFunction`, which only understands delays of less than 10 seconds.

## Motivation and Context

The `RetryAfterDelayFunction` incorrectly checks the input to be a single digit, rather than an integer number. That leads to the effect that any value that is greater than or equal to 10 will be ignored and retries will fallback to the configured delay.

While fixing this I ended up with an integration tests actually had to run 10 seconds to prove that the bug exists. I considered this too long, since riptide already takes a while to build. I refactored the code within the `RetryAfterDelayFunction` and tried to split it up into individual units, each having exactly one task:

- `RetryAfterDelayFunction`: Extract `Retry-After` header values from `HttpResponseException` 
- `CompoundDelayParser`: Delegate parsing of delays to specific parsers
- `SecondsDelayParser`: Parse delays expressed in seconds
- `HttpDateDelayParser`: Parse delays expressed as date times

I was now able to write specific tests for those individual units, which are quite a bit faster than the network-based integration tests.

## Types of changes

## Checklist:

### DIFF
--- a/riptide-failsafe/src/main/java/org/zalando/riptide/failsafe/CompoundDelayParser.java
+++ b/riptide-failsafe/src/main/java/org/zalando/riptide/failsafe/CompoundDelayParser.java
@@ -1,0 +1,33 @@
+package org.zalando.riptide.failsafe;
+
+import lombok.extern.slf4j.Slf4j;
+import net.jodah.failsafe.util.Duration;
+
+import java.util.Collection;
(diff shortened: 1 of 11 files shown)


lukasniemeier-zalando:
:+1:

PR #418, opened 2018-07-05
Added clientId tag to all metrics

### LINKED ISSUE #416: Add client id to all metrics tags
## Detailed Description

All metrics (request, retries and circuit breaker states) are all tagged with `clientName` which is used by spring and usually contains the hostname of the request URI. Since two clients could technically use the same base url and therefore host, it makes sense that the spring boot starter sets the client id as a default tag to all metrics recorders.

## Possible Implementation

Add `clientId` tag to:
- MetricsRetryListener
- MetricsCircuitBreakerListener
- MetricsPlugin
  - requires also to be registered per client, and not globally

## Your Environment

* Version used:
* Link to your project:

whiskeysierra:
## Motivation and Context

Fixes #416

## Types of changes

## Checklist:


lukasniemeier-zalando:
:+1:

PR #427, opened 2018-07-10
Fixed transitive transient fault detection

### LINKED ISSUE #426: Transient faults are not detected as transient themselves
## Description

In case http requests are made within the callback of another callback then the transient fault detection runs twice.

## Expected Behavior

`TransientFaultException` are wrapped in another.

## Actual Behavior

The plugin should treat `TransientFaultException`s already as transient and return them as-is.

## Possible Fix

1. Add `TransientFaultException` to list of default exception types marked as transient...
2. Or add it always, regardless of whether users build on top of the defaults or not

## Context

Wanted to change log level based on exception type and realized that we don't properly short-circuit there.

## Your Environment

* Version used: 2.8.1

whiskeysierra:
## Description

Transient fault exceptions are no longer wrapped in one another.

## Motivation and Context

Fixes #426

## Types of changes

## Checklist:


lukasniemeier-zalando:
:+1:

PR #428, opened 2018-07-10
Removed default circuit breaker

### LINKED ISSUE #423: Failsafe plugin uses a default circuit breaker instead of none
## Detailed Description

When creating a blank `FailsafePlugin` it uses a retry policy that *never* retries but a circuit breaker that opens after 1 failed request and without a delay immediately goes back to half open and then open state.

## Context

Users should reasonably expect that `new FailsafePlugin()` configures neither a retry policy nor a circuit breaker. That default circuit breaker is unexpected and not a good default for us.

## Possible Implementation

- Initialize circuit breaker property with `null`
- change `with(*)` call chain to a more functional approach
  - `private Failsafe withRetryPolicy(final Failsafe failsafe)`
  - `private Failsafe withCircuitBreaker(final Failsafe failsafe)`

whiskeysierra:
## Motivation and Context

Fixes #423

## Types of changes

## Checklist:


riptide-failsafe/src/main/java/org/zalando/riptide/failsafe/FailsafePlugin.java
```diff
@@ -29,11 +31,11 @@
     private final RetryListener listener;
 
     public FailsafePlugin(final ScheduledExecutorService scheduler) {
-        this(scheduler, NEVER, new CircuitBreaker(), RetryListener.DEFAULT);
+        this(scheduler, NEVER, null, RetryListener.DEFAULT);
```
AlexanderYastrebov: It is unfortunate that `CircuitBreaker` is not an interface and there is nothing like `CircuitBreaker.NOOP`

riptide-failsafe/src/main/java/org/zalando/riptide/failsafe/FailsafePlugin.java
```diff
@@ -55,9 +57,8 @@ public FailsafePlugin withListener(final RetryListener listener) {
     @Override
     public RequestExecution prepare(final RequestArguments arguments, final RequestExecution execution) {
         return () -> {
-            final CompletableFuture<ClientHttpResponse> original = Failsafe
-                    .with(retryPolicy)
-                    .with(circuitBreaker)
+            final CompletableFuture<ClientHttpResponse> original =
+                    withCircuitBreaker(with(retryPolicy))
```
AlexanderYastrebov: I would introduce 
```
private SyncFailsafe<Object> withRetries() {
  var failsafe = Failsafe.with(retryPolicy)
  return return circuitBreaker == null ? failsafe : failsafe.with(circuitBreaker);
}
```
instead

riptide-failsafe/src/test/java/org/zalando/riptide/failsafe/FailsafePluginWithoutCircuitBreakerTest.java
```diff
@@ -0,0 +1,91 @@
...
+        driver.addExpectation(onRequestTo("/foo"), giveEmptyResponse());
+
+        unit.get("/foo").call(pass())
+                .exceptionally(this::ignore)
+                .join();
+
+        final CompletableFuture<Void> timeout = unit.get("/foo").call(pass());
+        final CompletableFuture<Void> last = unit.get("/foo").call(pass());
+
+        timeout.exceptionally(partially(SocketTimeoutException.class, this::ignore)).join();
```
AlexanderYastrebov: I would inline these two the same way as the first call

riptide-failsafe/src/test/java/org/zalando/riptide/failsafe/RetryAfterDelayFunctionTest.java
```diff
@@ -62,11 +60,7 @@
                     .withRetryPolicy(new RetryPolicy()
                             .withDelay(2, SECONDS)
                             .withDelay(new RetryAfterDelayFunction(clock))
-                            .withMaxRetries(4))
-                    .withCircuitBreaker(new CircuitBreaker()
-                            .withFailureThreshold(3, 10)
-                            .withSuccessThreshold(5)
-                            .withDelay(1, TimeUnit.MINUTES)))
+                            .withMaxRetries(4)))
```
AlexanderYastrebov: This change did not break any test, didn't it?

riptide-failsafe/src/test/java/org/zalando/riptide/failsafe/FailsafePluginWithoutCircuitBreakerTest.java
whiskeysierra: I need to start both in parallel, that's why I create them separately and call join later.

riptide-failsafe/src/test/java/org/zalando/riptide/failsafe/RetryAfterDelayFunctionTest.java
whiskeysierra: Wasn't used here. Actually I'm missing a proper CircuitBreakerTest

riptide-failsafe/src/main/java/org/zalando/riptide/failsafe/FailsafePlugin.java
whiskeysierra: Kotlin?

whiskeysierra: Done

riptide-failsafe/src/test/java/org/zalando/riptide/failsafe/RetryAfterDelayFunctionTest.java
whiskeysierra: Added test

AlexanderYastrebov:
:+1:

PR #438, opened 2018-07-12
Introduced Metrics factory for starter

whiskeysierra:
## Description

Introduced static factory for everything metrics related, i.e. plugin and listeners.

## Motivation and Context

Right now we used internal constructors since Spring doesn't support `with*` methods. That forced us to repeat the default for the metrics name parameters in three places which we solved by exposing a constant everywhere. 

By introducing a static factory those downsides can now all be avoided.

## Types of changes

## Checklist:


lukasniemeier-zalando:
:+1:

PR #439, opened 2018-07-12
Fixed incorrect scheduler pool size

whiskeysierra:
## Description

The size of the scheduler used for backup requests, retries and circuit breaker is no longer initialized with 1.

## Motivation and Context

That scheduler basically was a bottleneck in the best case (concurrent requests being serialized) and a full-blown deadlock in the worst case (nested requests waited forever).

## Types of changes

## Checklist:


lukasniemeier-zalando:
:+1:

PR #447, opened 2018-07-14
Optimized usage of Plugin.compound(..)

### LINKED ISSUE #446: Plugin.compound(..) is called upon every request
## Description

Instead of once at construction time, we combine all plugins into one on every request

## Expected Behavior

Plugins are combined once into a compound plugin.

## Actual Behavior

Happens on every request.

## Possible Fix

- [ ] Move `Plugin.compound(..)` from `Requester` into `DefaultHttpBuilder`

## Context

We should try to avoid creating unnecessary objects.

## Your Environment

* Version used: 2.9.0-RC.4

whiskeysierra:
## Motivation and Context

Fixes #446

## Types of changes

## Checklist:


riptide-core/src/main/java/org/zalando/riptide/CompoundPlugin.java
```diff
@@ -1,23 +1,35 @@
...
 final class CompoundPlugin implements Plugin {
 
-    private final Plugin left;
-    private final Plugin right;
+    private final Collection<Plugin> plugins;
 
-    public CompoundPlugin(final Plugin left, final Plugin right) {
-        this.left = left;
-        this.right = right;
+    CompoundPlugin(final Collection<Plugin> plugins) {
```
AlexanderYastrebov: Compound or Composite? https://en.wikipedia.org/wiki/Composite_pattern

AlexanderYastrebov:
:+1:

riptide-core/src/main/java/org/zalando/riptide/CompoundPlugin.java
whiskeysierra: Until now I've consistently used *Compound*.

PR #453, opened 2018-07-16
Removed synchronous exception, now it's always a future

### LINKED ISSUE #449: Always return CompletableFuture
## Detailed Description

Instead of throwing an `UncheckedIOException` in `ResponseDispatcher.call(Route)` we may wrap any exception in a `CompletableFuture` and return that one instead. That removes the need to wrap the whole call in a `try-catch` and unifies the exception handling approach (by allowing `CompletableFuture.exceptionally(Function)` for any kind of exception).

## Context

This issue is part of a bigger initiative for 3.x to propose a better, more powerful exception handling mechanism.

This is actually just a preliminary step towards an even better solution. Right now the only thing that happens on the IO thread is the waiting for the response and the response processing. Everything related to the request, including but not limited to DNS lookup, body serialization, opening a connection and writing the request, is done on the calling thread. All of that has to move to the IO thread already. Unfortunately this is not straightfoward with the current API of `AsyncClientHttpRequestFactory` and `AsyncClientHttpRequest`.

## Possible Implementation

Replace `throw` with a new `CompletableFuture` here:
https://github.com/zalando/riptide/blob/830ec392182e5a9e09ead8344a36dd7f3b3c952e/riptide-core/src/main/java/org/zalando/riptide/Requester.java#L162

## Your Environment

* Version used: 2.9.0-RC.4

whiskeysierra:
## Motivation and Context

Fixes #449

## Types of changes

## Checklist:

### DIFF
--- a/riptide-core/src/main/java/org/zalando/riptide/Requester.java
+++ b/riptide-core/src/main/java/org/zalando/riptide/Requester.java
@@ -16,7 +16,6 @@ import org.zalando.fauxpas.ThrowingUnaryOperator;
 
 import javax.annotation.Nullable;
 import java.io.IOException;
-import java.io.UncheckedIOException;
 import java.net.URI;
 import java.time.OffsetDateTime;
 import java.util.Arrays;
@@ -122,13 +121,13 @@ public final class Requester extends Dispatcher {
         return copy;
     }
 
-    public <T> Dispatcher body(@Nullable final T body) {
-        return execute(body);
-    }
-
     @Override
     public CompletableFuture<ClientHttpResponse> call(final Route route) {
-        return execute(null).call(route);
+        return body(null).call(route);
+    }
+
+    public <T> Dispatcher body(@Nullable final T body) {
+        return execute(body);
     }
 
     private <T> Dispatcher execute(final @Nullable T body) {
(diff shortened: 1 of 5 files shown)


riptide-core/src/main/java/org/zalando/riptide/Requester.java
```diff
@@ -161,7 +160,10 @@ private HttpHeaders copyHeaders() {
 
                 return after.execute(arguments);
             } catch (final IOException e) {
```
tkrop: What about the unchecked exceptions?

whiskeysierra: Mmh. Good point.

riptide-core/src/main/java/org/zalando/riptide/Requester.java
```diff
@@ -160,8 +159,11 @@ private HttpHeaders copyHeaders() {
                 final RequestExecution after = plugin.beforeDispatch(dispatch(before, route));
 
                 return after.execute(arguments);
-            } catch (final IOException e) {
-                throw new UncheckedIOException(e);
+            } catch (final Exception e) {
+                // for issues that occur before the future was successfully created, i.e. request being sent
```
tkrop: Sounds misleading, and I'm not sure that you really need this.

tkrop:
:+1:

PR #455, opened 2018-07-16
Failsafe plugin now fails to compile if used incorrectly

### LINKED ISSUE #442: FailsafePlugin's API should enforce correct usage
## Detailed Description

Using the public no-argument constructor without one of the `with*` methods should be impossible.

## Context

The API/compiler can enforce correct usage and replace a runtime issue with a compile time error.

## Possible Implementation

- FailsafePlugin should no longer implement `Plugin` directly
- only `withRetryPolicy`/`withCircuitBreaker` returns an public static inner class (or interface) type that is in fact a plugin

whiskeysierra:
Fixes #442

## Motivation and Context

Fixes #442 

## Types of changes

## Checklist:


tkrop:
:+1:

PR #460, opened 2018-07-18
Moved executor from request factory up to core

### LINKED ISSUE #457: Move request handling to IO thread
## Detailed Description

Instead of using the current thread we should use the IO thread already during request creating, i.e. earlier.

## Context

Right now the only thing that happens on the IO thread is the waiting for the response and the response processing. Everything related to the request, including but not limited to DNS lookup, body serialization, opening a connection and writing the request, is done on the calling thread. All of that has to move to the IO thread already. Unfortunately this is not straightfoward with the current API of AsyncClientHttpRequestFactory and AsyncClientHttpRequest.

Related to #449.

## Possible Implementation

1. **Move from `AsyncClientHttpRequestFactory` to `ClientHttpRequestFactory`**
   - [x] Modify `RestAsyncClientHttpRequestFactory` 
     - [x] to not implement `AsyncClientHttpRequestFactory`
     - [x] remove `AsyncListenableTaskExecutor` dependency
   - [x] Maybe take the opportunity to rename everything in `riptide-httpclient` to `Apache*`?
   - [x] Modify `Http`'s builder API
     - [x] Add new stage `ExecutorStage` to `Http` interface which requires an `Executor`
     - [x] Make new stage mandatory first or second step when constructing `Http` instance
     - [x] Remove `RequestFactoryStage.simpleRequestFactory(..)`
   - [x] Pass executor to `DefaultHttp`
   - [x] Pass executor to `Requester`
   - [x] Call the following methods within a runnable passed to the executor
     - [x] `ClientHttpRequestFactory.createRequest(..)`
     - [x] `MessageWorker.write(..)`
     - [x] `ClientHttpRequest.execute()`
   - [ ] Enabled warning on deprecation now? See #329 

## Your Environment

* Version used: 2.9.0-RC.4

whiskeysierra:
## Motivation and Context

Fixes #457

## Types of changes

## Checklist:


whiskeysierra:
:+1:

PR #491, opened 2018-09-18
bugfix/add-spring2.x-place-holder-resolver

### LINKED ISSUE #488: SpringBoot2xSettingsParser fails to resolve placeholders
## Description
When using riptide 2.8.1, and defining scopes as follows 
```yaml
riptide:
  clients:
    user-detail:
      oauth.scopes: uid, ${SCOPE:service.read}
```
The scopes are read as is without being resolved!
With a bit of debugging it seems this line(https://github.com/zalando/riptide/blob/7ee3f0917b213303d93d36fdb9c0cba524897643/riptide-spring-boot-2.x-support/src/main/java/org/zalando/riptide/spring/SpringBoot2xSettingsParser.java#L24)
It constructs a binder with no PlaceHolderResolver.

mazenmelouk:
adds a placeholder resolver to spring 2.x binder
fixes #488

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
     <groupId>org.zalando</groupId>
     <artifactId>riptide-parent</artifactId>
-    <version>2.9.0-SNAPSHOT</version>
+    <version>2.10.0-SNAPSHOT</version>
     <packaging>pom</packaging>
 
     <name>Riptide</name>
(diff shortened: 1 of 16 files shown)


whiskeysierra:
Can you add a test?

whiskeysierra:
I believe you're missing an import somewhere. Also, does this need to be fixed for spring boot 1.x as well?

whiskeysierra:
https://travis-ci.org/zalando/riptide/jobs/430404676

mazenmelouk:
okai, do you see any other way of fixing it? Rather that having Spring boot 2 as dependency

whiskeysierra:
Replaced by #499

PR #507, opened 2018-10-23
test encoding

patandrick:
test for expected encoding of datetime (will fail)

## Description
We had unexpected issues with datetimes like 
2018-05-21T10:24:47.788+02:00 which are not encoded for the url as a query parameter.
If you encode manually before the request (2018-05-21T10%3A24%3A47.788%2B02%3A00)
then this will be encoded again to "2018-05-21T10%253A24%253A47.788%252B02%253A00"

### DIFF
--- a/riptide-core/src/main/java/org/zalando/riptide/RequestArguments.java
+++ b/riptide-core/src/main/java/org/zalando/riptide/RequestArguments.java
@@ -6,10 +6,12 @@ import org.apiguardian.api.API;
 import org.springframework.http.HttpMethod;
 import org.springframework.util.MultiValueMap;
 import org.springframework.web.util.UriComponentsBuilder;
+import org.springframework.web.util.UriUtils;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.net.URI;
+import java.nio.charset.Charset;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static org.apiguardian.api.API.Status.STABLE;
@@ -95,8 +97,8 @@ public interface RequestArguments {
         {
             final UriComponentsBuilder components = UriComponentsBuilder.newInstance();
             getQueryParams().entries().forEach(entry ->
-                    components.queryParam(entry.getKey(), entry.getValue()));
-            queryParams = components.build().encode().getQueryParams();
+                    components.queryParam(entry.getKey(), UriUtils.encode(entry.getValue(), Charset.defaultCharset())));
+            queryParams = components.build().getQueryParams();
         }
 
         // build request uri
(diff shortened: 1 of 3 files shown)


whiskeysierra:
https://jira.spring.io/browse/SPR-16860?focusedCommentId=160783&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-160783 might be a workaround for us

riptide-core/src/main/java/org/zalando/riptide/RequestArguments.java
```diff
@@ -95,8 +97,8 @@ default RequestArguments withRequestUri() {
         {
             final UriComponentsBuilder components = UriComponentsBuilder.newInstance();
             getQueryParams().entries().forEach(entry ->
-                    components.queryParam(entry.getKey(), entry.getValue()));
-            queryParams = components.build().encode().getQueryParams();
+                    components.queryParam(entry.getKey(), UriUtils.encode(entry.getValue(), Charset.defaultCharset())));
```
whiskeysierra: That won't compile in Spring 4, I believe since the overloaded version with `Charset` didn't exist back then.

This should work:

```java
getQueryParams().entries().forEach(throwingConsumer(entry ->
        components.queryParam(entry.getKey(), encode(entry.getValue(), "UTF-8"))));
```

riptide-core/src/main/java/org/zalando/riptide/RequestArguments.java
```diff
@@ -95,8 +97,8 @@ default RequestArguments withRequestUri() {
...
+            queryParams = components.build().getQueryParams();
```
whiskeysierra: Shouldn't this be `build(true)`, since it's already encoded?

riptide-core/src/test/java/org/zalando/riptide/RequesterTest.java
```diff
@@ -114,6 +114,28 @@ public void shouldEncodeAppendedQueryParams() {
                         on(SUCCESSFUL).call(pass()));
     }
 
+    @Test
+    public void shouldAppendedDateTimeQueryParams() {
+        expectRequestTo("https://test.datetimes.org/index.php?to=2018-05-21T10%3A24%3A47.788&from=2017-04-20T09%3A23%3A46.787");
+
+        unit.head("https://test.datetimes.org/index.php")
+                .queryParam("to", "2018-05-21T10:24:47.788")
```
whiskeysierra: Not relevant for the test, but proper ISO 8601 has no milliseconds and it has to have `Z` for UTC or an offset.

patandrick:
:+1:

whiskeysierra:
:+1:

patandrick:
:+1:

PR #527, opened 2018-11-16
Moved string converter to last place

### LINKED ISSUE #490: String Converter takes precedence over JSON
## Description

If a client wants to read a JSON value as a string (`on(OK).call(String.class, ...)`) the `StringHttpMessageConverter` is used instead.

## Expected Behavior

Jackson should be used to deserialize.

## Actual Behavior

String converter takes precedence (since it supports every media type, supports String and comes first).
It fails since the string converter falls back to ISO charset instead of UTF-8.

## Possible Fix

Move String converter after jackson.

## Steps to Reproduce

1. Read `application/json` response from server that includes unicode characters.
2. Read it as a String

## Context

Keep original JSON without modification.

whiskeysierra:
## Motivation and Context

Fixes #490

## Types of changes

## Checklist:


whiskeysierra:
:+1:

PR #579, opened 2019-02-07
Added ability to release connection instead of abort

### LINKED ISSUE #519: Release connection instead of abort
## Detailed Description

Within the `RestAsyncClientHttpResponse` we abort connections when closing the response instead of releasing it back to the pool.

## Context

Releasing it back to the pool would in theory allow for a connection to be reused.

## Possible Implementation

https://github.com/zalando/riptide/blob/0b8928a07b212b732590f4b9aaccdc8056ab8d20/riptide-httpclient/src/main/java/org/zalando/riptide/httpclient/RestAsyncClientHttpResponse.java#L43

Call [`releaseConnection`](https://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/conn/ConnectionReleaseTrigger.html#releaseConnection()) instead.

## Your Environment

* Version used: 2.10.1

### LINKED ISSUE #520: Not all routes close responses
## Description

Not all `Route` implementations are properly closing and therefore releasing their connection. See also #519.

## Expected Behavior

All routes should close their response.

## Actual Behavior

At least `PassRoute` and `NoRoute` don't close responses.

## Possible Fix

1. Make *closing the response* part of the contract of a `Route` and change every implementation. That's a potential breaking change since the semantic of the interface changed.
2. Handle closing of responses in a central place somewhere. Not sure whether it's possible since we have cases where the closing is actively delayed, e.g. when returning an `AutoCloseable` response entity e.g. a `Stream`.

## Steps to Reproduce

Not 100% but worth having a test for this:

1. Configure connection pool size of 1 (is that even possible? what's the default?)
2. Make one request and pass on the respnse
3. Do another request and pass on the response
4. 💥 Should timeout while waiting for connection now

## Context

We're investigating an exhausted connection pool and tried to identify reasons how connections might get stale.

## Your Environment

* Version used: 2.2.2 (findings might not apply to that old version though)

whiskeysierra:
Fixes #519
Fixes #520

## Types of changes

## Checklist:

### DIFF
--- a/riptide-httpclient/src/main/java/org/zalando/riptide/httpclient/EndOfStreamDetectingInputStream.java
+++ b/riptide-httpclient/src/main/java/org/zalando/riptide/httpclient/EndOfStreamDetectingInputStream.java
@@ -1,0 +1,54 @@
+package org.zalando.riptide.httpclient;
+
+import org.zalando.fauxpas.ThrowingBiConsumer;
+
+import java.io.FilterInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+final class EndOfStreamDetectingInputStream extends FilterInputStream {
+
+    private boolean ended;
+    private final ThrowingBiConsumer<InputStream, Boolean, IOException> closer;
+
+    EndOfStreamDetectingInputStream(final InputStream in,
+            final ThrowingBiConsumer<InputStream, Boolean, IOException> closer) {
+        super(in);
+        this.closer = closer;
+    }
+
+    @Override
+    public int read() throws IOException {
+        final int read = super.read();
+
+        if (read == -1) {
+            ended = true;
+        }
+
+        return read;
+    }
+
+    @Override
+    public int read(byte[] b) throws IOException {
+        return onRead(super.read(b), b.length);
+    }
+
+    @Override
+    public int read(byte[] b, int off, int len) throws IOException {
+        return onRead(super.read(b, off, len), len);
+    }
+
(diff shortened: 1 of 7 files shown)


riptide-httpclient/src/main/java/org/zalando/riptide/httpclient/EndOfStreamDetectingInputStream.java
```diff
@@ -0,0 +1,54 @@
...
+        return onRead(super.read(b), b.length);
+    }
+
+    @Override
+    public int read(byte[] b, int off, int len) throws IOException {
+        return onRead(super.read(b, off, len), len);
+    }
+
+    private int onRead(final int read, final int max) {
+        if (read < max) {
```
jhorstmann: Is this guaranteed to indicate the end of the stream? I would have guessed that the number of bytes read depends on the underlying network buffers. Or is apache http client blocking until the byte array can be filled fully?

whiskeysierra: Good point. I was just assuming it's reliable. I'll change it to check for `-1`, that will definitely work.

whiskeysierra: Fixed in https://github.com/zalando/riptide/pull/580

PR #584, opened 2019-02-13
Replaced tokens library with internal K8s-only implementation

whiskeysierra:
## Description

Replaces the overloaded tokens library with a k8s-only implementation.

## Motivation and Context

The tokens library was useful for STUPS applications, but is pretty much useless for K8s.

## Types of changes

## Checklist:


whiskeysierra:
@lukasniemeier-zalando I dropped the `scopes` from the `oauth` namespace, since this is defined within the credentials.yaml for CDP. In order to have a non-empty `oauth` namespace I opted for a pretty dumb `enabled: true`. I don't like the mismatch between this and let's say circuit breaker and retry which become active the moment that key appears. Any other ideas?

whiskeysierra:
Although an `enabled` property everywhere would allow users to specify defaults for optional features without immediately turning them on for every client... :thinking:

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/DefaultRiptideRegistrar.java
```diff
@@ -227,85 +235,126 @@ private String findObjectMapper(final String id) {
...
+            return Optional.of(pluginId);
         }
+        return Optional.empty();
+    }
 
+    private Optional<String> registerTransientFaultPlugin(final String id, final Client client) {
         if (client.getDetectTransientFaults()) {
             log.debug("Client [{}]: Registering [{}]", id, TransientFaultPlugin.class.getSimpleName());
-            plugins.add(registry.registerIfAbsent(id, TransientFaultPlugin.class, () ->
+            final String e = registry.registerIfAbsent(id, TransientFaultPlugin.class, () ->
```
lukasniemeier-zalando: `e` ? :-D

whiskeysierra:
@lukasniemeier-zalando Almost done. I'm missing a good name for the `timeout` plugin section.

riptide-auth/src/main/java/org/zalando/riptide/auth/PlatformCredentialsAuthorizationProvider.java
```diff
@@ -0,0 +1,56 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.List;
+
+import static com.google.common.base.Preconditions.checkElementIndex;
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+/**
+ * A special {@link AuthorizationProvider authorization provider} that built for Zalando's Platform IAM which provides
```
lukasniemeier-zalando: the `that` is floating in mid-air it seems

lukasniemeier-zalando:
:+1:

Issue #509, opened 2018-10-28
OpenTracing support


whiskeysierra:
## Detailed Description

Add OpenTracing integration that e.g. allows to add retry count as a tag to spans.

## Context

Instrumenting the underlying Apache Http Client is a good enough solution for most, but a deeper integration allows for more sophisticated use cases (like retry count).

## Possible Implementation

- [x] `riptide-opentracing` module
  - [x] `OpenTracingPlugin` 
  - [x] plugin that creates a span for one logical request
  - [x] `OpenTracingRetryListener` which set's a retry count tag
  - [x] Support for span decorator?
    - [x] Tags https://github.com/opentracing/specification/blob/master/semantic_conventions.md
  - [x] Operation Name generator?
    - [x] Type-safe attributes to be passed from call-site to plugins (transported as part of request arguments)
- [x] `riptide-spring-boot-autoconfigure`
  - [x] plugin needs to be registered directly after/before timeout plugin (auto configuration)
  - [x] instrument executor and scheduler using [opentracing-concurrent](https://github.com/opentracing-contrib/java-concurrent)
  - [x] instrument Apache Http Client using [jaeger-apachehttpclient](https://github.com/jaegertracing/legacy-client-java/tree/master/jaeger-apachehttpclient) without dependency to jaeger!
    - [x] Propagate tags?
  - [x] Set client id as tag?
  - [x] Degree of configurability?

### Example Configuration

```yaml
riptide:
  defaults:
    tracing:
      enabled: true
      tags:
        account: ${CDP_TARGET_INFRASTRUCTURE_ACCOUNT}
        zone: ${CDP_TARGET_REGION}
        artifact_version: ${CDP_BUILD_VERSION}
        deployment_id: ${CDP_DEPLOYMENT_ID}
      propagate-flow-id: true
        
  clients:
    exchange-rate:
      tracing:
        tags:
          peer.service: exchange-rate-service
        propagate-flow-id: false
```

## Your Environment

* Version used:
* Link to your project:

tkrop:
I'm not sure about the value of retry count - OpenTracing needs aggregate over spans anyhow, making a retry count optical visible. However, a consistent higher level *operation name* and an *idempotency key* to distinguish different requests could be a value. Riptide does not provide an operation name, does it? Anyhow, it could be easy to calculate a consistent operation name from the path template.

tkrop:
> See also https://github.bus.zalan.do/SRE/opentracing/issues/6.

[...]

PR #595, opened 2019-02-24
OpenTracing support

whiskeysierra:
## Description

Added the ability to manage client span(s), decorate spans and inject the span context into requests.

## Motivation and Context

#509 

## Types of changes

## Checklist:


whiskeysierra:
@pc-alves @lmineiro Feedback is highly appreciated! :mag_right:

whiskeysierra:
Merge is blocked by https://github.com/zalando/tracer/pull/204 which in turn requires https://github.com/opentracing/opentracing-java/issues/314 :hourglass:

riptide-opentracing/src/main/java/org/zalando/riptide/opentracing/span/ErrorSpanDecorator.java
```diff
@@ -0,0 +1,26 @@
...
+import io.opentracing.Span;
+import io.opentracing.log.Fields;
+import io.opentracing.tag.Tags;
+import org.zalando.riptide.RequestArguments;
+
+/**
+ * Sets the <code>error</code> span tag as well as the <code>error.kind</code> and <code>error.object</code> span logs.
+ *
+ * @see <a href="https://opentracing.io/specification/conventions/#span-tags-table">Standard Span Tags</a>
+ * @see <a href="https://opentracing.io/specification/conventions/#log-fields-table>Standard Log Fields</a>
```
lukasniemeier-zalando: missing closing `"`

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/DefaultRiptideRegistrar.java
```diff
@@ -240,9 +250,11 @@ private String findObjectMapper(final String id) {
         final Stream<Optional<String>> plugins = Stream.of(
                 registerMetricsPlugin(id, client),
                 registerTransientFaultPlugin(id, client),
+                registerOpenTracingRetryPlugin(id, client),
```
lukasniemeier-zalando: Help me. Isn't this covered by `registerOpenTracingPlugin` (and check against failsafe retry in `OpenTracingPluginFactory`)?

whiskeysierra: If you have both `tracing.enabled: true` and `retry.enabled: true` then we want two things to happen:

1. The span should be decorated with `retry` tag and `retry_number`.
2. One additional enclosing client span for all retries.

Number 1 is achieved by registering the `RetrySpanDecorator` within the factory. Registering the very same plugin twice fixes number 2.

Having said that...

I debugged it a little bit to figure out how the spans will look like and I was hoping for this:

```
         [Span A] (enclosing)
             |
    +--------+--------+
    |                 |
[Span B] (first)  [Span C] (retry)
```

Unfortunately what we get is this:

```
[Span A] (enclosing)
    |
[Span B] (first)
    |
[Span C] (retry)
```

From what I understand this is due to how Failsafe triggers retries. Basically every failed attempt will schedule the retry, so the parent-child relationship is from one attempt to the next. When we would incorporate time into this it should look like a waterfall with varying delays in between (depending on backoff, jitter, etc).

whiskeysierra: I'd actually argue that we don't need the enclosing span in that case. Or we don't rely on the implicit *active-span-becomes-parent* and pass the original, enclosing span explicitly. I'm not 100% sure whether that's worth the effort. I guess the causal chain of retries should be enough.

@lmineiro Your opinion?

whiskeysierra: Meaning we could actually just get rid of the second plugin registration.

riptide-opentracing/README.md
```diff
@@ -0,0 +1,185 @@
...
+```
+
+### Span Decorators
+
+Span decorators are a simple, yet powerful tool to manipulate the span, i.e. they allow you to
+add tags, logs and baggage to spans. The default set of decorators can be extended by using 
+`OpenTracingPlugin#withAdditionalSpanDecorators(..)`:
+
+```java
+new OpenTracingPlguin(tracer)
```
lmineiro: typo: s/OpenTracingPlguin/OpenTracingPlugin

riptide-opentracing/README.md
```diff
@@ -0,0 +1,185 @@
...
+**Be aware**: The `http.url` tag is disabled by default because the full request URI may contain
+sensitive, [*personal data*](https://en.wikipedia.org/wiki/General_Data_Protection_Regulation).
+As an alternative we introduced the `http.path` tag which favors the URI template over the
+already expanded version. That has the additional benefit of a significant lower cardinality
+compared to what `http.url` would provide. 
+
+If you still want to enable it, you can do so by just registering the missing span decorator:
+
+```java
+new OpenTracingPlguin(tracer)
```
lmineiro: typo: s/OpenTracingPlguin/OpenTracingPlugin

riptide-opentracing/README.md
```diff
@@ -0,0 +1,185 @@
...
+    <version>${riptide.version}</version>
+</dependency>
+```
+
+## Configuration
+
+```java
+Http.builder()
+    .baseUrl("https://www.example.com")
+    .plugin(new OpenTracingPlguin(tracer)
```
lmineiro: typo: s/OpenTracingPlguin/OpenTracingPlugin

riptide-opentracing/README.md
```diff
@@ -0,0 +1,185 @@
...
+    .withAdditionalSpanDecorators(new StaticSpanDecorator(singletonMap(
+            "environment", "local"
+    )))
+```
+
+If the default span decorators are not desired you can replace them completely using
+`OpenTracingPlugin#withSpanDecorators(..)`:
+
+```java
+new OpenTracingPlguin(tracer)
```
lmineiro: typo: s/OpenTracingPlguin/OpenTracingPlugin

riptide-opentracing/README.md
```diff
@@ -0,0 +1,185 @@
...
+```java
+http.get("/users/{id}", userId)
+    .dispatch(series(),
+        on(SUCCESSFUL).call(User.class, this::greet),
+        anySeries().call(problemHandling()))
+```
+
+### Operation Name
+
+By default the HTTP method will be used as the operation name, which might not fit your needs.
```
lmineiro: Have you considered making the operation name more of a first-class citizen and an argument for the method `get()` (in the example)?

riptide-opentracing/src/main/java/org/zalando/riptide/opentracing/span/UriVariablesTagSpanDecorator.java
```diff
@@ -0,0 +1,56 @@
...
+        @Nullable final String template = arguments.getUriTemplate();
+
+        if (template == null) {
+            return ImmutableMap.of();
+        }
+
+        return extract(template, arguments.getUriVariables());
+    }
+
+    @Hack("Pretty dirty, but I couldn't find any other way...")
```
lmineiro: 🤣

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/DefaultRiptideRegistrar.java
lmineiro: I agree that it would be reasonable not to force the enclosing span complexity. It felt slightly dirty anyway

whiskeysierra: Agreed. I'll get rid of the second plugin registration. We still have that option later.

riptide-opentracing/README.md
whiskeysierra: That would break our modularization concept and would require OpenTracing to be part of the core rather than a separate add-on module.

riptide-opentracing/src/main/java/org/zalando/riptide/opentracing/span/ErrorSpanDecorator.java
whiskeysierra: Fixed

riptide-opentracing/README.md
```diff
@@ -0,0 +1,185 @@
...
+```java
+Http.builder()
+    .plugin(new OpenTracingPlugin(tracer))
+    .build();
+```
+
+## Features
+
+- Client span lifecycle management
+- Span injection into HTTP headers of requests
```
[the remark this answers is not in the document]
whiskeysierra: Fixed

riptide-opentracing/README.md
lmineiro: Understood. Sounds reasonable.
In a separate initiative (universal RPC), we're thinking about enforcing naming the operation. This already seemed very close to what we've come up with so far, that's why I was asking

riptide-opentracing/README.md
```diff
@@ -0,0 +1,185 @@
...
+    <version>${riptide.version}</version>
+</dependency>
+```
+
+## Configuration
+
+```java
+Http.builder()
+    .baseUrl("https://www.example.com")
+    .plugin(new OpenTracingPlugin(tracer)
```
pc-alves: Missing `)`

riptide-opentracing/README.md
```diff
@@ -0,0 +1,185 @@
...
+            new HttpUrlSpanDecorator(),
+            new HttpStatusCodeSpanDecorator(),
+            new ErrorSpanDecorator(),
+            new CallSiteSpanDecorator())
+```
+
+## Usage
+
+Typically you won't need to do anything at the call-site regarding OpenTracing, i.e.
+you usages of Riptide should work exactly as before:
```
pc-alves: s/you usages/your usages/

whiskeysierra: Fixed

riptide-opentracing/README.md
whiskeysierra: Fixed

whiskeysierra:
I should add a test that verifies that an error response that is retried and eventually succeeds is not counted as an error. The first attempt should though.

whiskeysierra:
For retries we might want to use follows-from rather than child-of. @lmineiro @pc-alves Your thoughts?

pc-alves:
Retries should still use the `childOf` relationship type. The guideline is that `childOf` should be used when "the parent Span depends on the child Span in some capacity" ([OpenTracing Specification - References between Spans](https://opentracing.io/specification/)). This still holds true for retries.

whiskeysierra:
Alright, that works then. Thanks for the info!

riptide-opentracing/README.md
```diff
@@ -0,0 +1,185 @@
+# Riptide: OpenTracing
+
+[![Spider web](../docs/spider-web.jpg)](https://pixabay.com/photos/cobweb-drip-water-mirroring-blue-3725540/)
+
+[![Build Status](https://img.shields.io/travis/zalando/riptide/master.svg)](https://travis-ci.org/zalando/riptide)
+[![Coverage Status](https://img.shields.io/coveralls/zalando/riptide/master.svg)](https://coveralls.io/r/zalando/riptide)
+[![Code Quality](https://img.shields.io/codacy/grade/1fbe3d16ca544c0c8589692632d114de/master.svg)](https://www.codacy.com/app/whiskeysierra/riptide)
+[![Javadoc](https://www.javadoc.io/badge/org.zalando/riptide-metrics.svg)](http://www.javadoc.io/doc/org.zalando/riptide-metrics)
```
lukasniemeier-zalando: copy-pasta

lukasniemeier-zalando:
:+1:

PR #598, opened 2019-03-01
Added Basic Auth support

whiskeysierra:
## Description

Added a `BasicAuthorizationProvider` in addition to the one we already have which is specific for Zalando's Platform IAM.

## Motivation and Context

Until now the auth plugin is rather limited for non-Zalando users.

## Types of changes

## Checklist:

### DIFF
--- a/riptide-auth/src/main/java/org/zalando/riptide/auth/BasicAuthorizationProvider.java
+++ b/riptide-auth/src/main/java/org/zalando/riptide/auth/BasicAuthorizationProvider.java
@@ -11,25 +11,22 @@ import static java.nio.charset.StandardCharsets.ISO_8859_1;
  */
 public final class BasicAuthorizationProvider implements AuthorizationProvider {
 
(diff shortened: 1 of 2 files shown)


riptide-auth/src/main/java/org/zalando/riptide/auth/BasicAuthorizationProvider.java
```diff
@@ -0,0 +1,35 @@
...
+        checkArgument(encoder.canEncode(password), "Password must be encoded in ISO-8859-1");
+        this.username = username;
+        this.password = password;
+    }
+
+    @Override
+    public String get() {
+        final String credentials = username + ":" + password;
+        final byte[] bytes = credentials.getBytes(ISO_8859_1);
+        return "Basic " + base64.encodeToString(bytes);
```
lukasniemeier-zalando: Move to ctx? Think about the CPU cycles ;-p

lukasniemeier-zalando:
:+1:

PR #602, opened 2019-03-01
Introduced persistent collections

### LINKED ISSUE #600: Don't expose Immutable* data types
## Detailed Description

Especially `RequestArguments` should not expose e.g. `ImmutableMultimap*` but rather a plain
`Map<String, List<String>>`. Applies to headers, query, URI variables and attributes.

## Context

We need a bit more freedom to choose different implementations under the hood. As an example: https://pcollections.org

## Possible Implementation

Change all usages from:
- [ ] `ImmutableMultimap` to `Map<String, List<String>>`
- [ ] `ImmutableMap<String, String>` to `Map<String, String>`
- [ ] `ImmutableList<String>` to `List<String>`

### LINKED ISSUE #601: Use persistent data structures for request arguments
## Detailed Description

Request arguments require to make copies of maps/lists even for just adding a single new entry, e.g. a header. 

## Context

A persistent data structure would be more efficient.

Requires #600.

## Possible Implementation

- [ ] https://pcollections.org
- [ ] Only expose standard Java Collection interfaces
- [ ] `withHeader(String, String)`
- [ ] `withHeaders(Map<String, List<String>>)`
- [ ] `replaceHeaders(Map<String, List<String>>)`
- [ ] Same for query parameters and URI variables
- [ ] Attributes should only have the singular version

whiskeysierra:
## Motivation and Context

Fixes #600 
Fixes #601 

## Types of changes

## Checklist:

### DIFF
--- a/riptide-core/pom.xml
+++ b/riptide-core/pom.xml
@@ -25,6 +25,11 @@
             <artifactId>spring-context</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.organicdesign</groupId>
+            <artifactId>Paguro</artifactId>
+            <version>3.1.2</version>
+        </dependency>
+        <dependency>
             <groupId>org.springframework</groupId>
             <artifactId>spring-test</artifactId>
         </dependency>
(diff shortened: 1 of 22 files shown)


riptide-core/pom.xml
```diff
@@ -24,6 +24,11 @@
             <groupId>org.springframework</groupId>
             <artifactId>spring-context</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.organicdesign</groupId>
+            <artifactId>Paguro</artifactId>
```
lukasniemeier-zalando: Pagurus is a genus of hermit crabs in the family Paguridae.

lukasniemeier-zalando:
:+1:

PR #606, opened 2019-03-02
Fixed caching test

### LINKED ISSUE #605: Force caching
## Detailed Description

Not every server is providing proper cache headers, yet clients often know that a certain resource can be cached. Riptide should come with the option to force caching, even without the server's *consent*.

## Context

Right now users are forced to implement their own cache within their gateway or business logic when it could be done very transparent within Riptide.

https://opensource.zalando.com/restful-api-guidelines/#227

## Possible Implementation

Maybe a custom interceptor that modifies the response before the Http Client inspects it would be hacky way to do it.

whiskeysierra:
## Description

Turns out heuristic caching works as expected, the problem was that the test setup was using empty responses with status code 204 which are not considered cacheable.

## Motivation and Context

Fixes #605

## Types of changes

## Checklist:


lukasniemeier-zalando:
:+1:

PR #607, opened 2019-03-02
Switched from compound to composite

whiskeysierra:
## Motivation and Context

Fixes #599 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.

Issue #604, opened 2019-03-02
Native Logbook support


whiskeysierra:
## Detailed Description

Introduce a new module and plugin implementation to add agnostic logbook support.

## Context

Independent of the underlying HTTP client library.

## Possible Implementation

- [x] Requires #603 
- [ ] Requires zalando/logbook#441
- [ ] Requires Logbook 2.x release
- [ ] New `riptide-logbook` module
- [ ] New `LogbookPlugin`, binding to the *serialization* phase
- [ ] Replace dependency to `LogbookHttpRequestInterceptor` with `LogbookPlugin`
- [ ] Plugin can be shared across clients

PR #609, opened 2019-03-02
Added serialization phase

### LINKED ISSUE #603: New Plugin Phase: Serialization
## Detailed Description

Introduce a new plugin phase *serialization* (or *request-serialization*?) to allow plugins to post-process serialized request bodies.

## Context

Enables the following use cases:
- #608 `RequestCompressionPlugin` (rather than an interceptor that is specific to Apache HTTP Client)
- `LogbookPlugin` (rather than an interceptor)
- Allows us to properly implement the plugin interceptor
  - Pass serialized body to the right place

https://github.com/zalando/riptide/blob/ad63d736a1daff1690eed1e496b8c39c43a887fc/riptide-core/src/main/java/org/zalando/riptide/PluginInterceptor.java#L118-L130

## Possible Implementation

- [x] Phase *serialization*, `Plugin.aroundSerialization`
- [x] After *dispatch*, but before *network*
- [x] Implementation: Delegate to `MessageWriter`
- [x] Keep original and serialized body in `RequestArguments`
  - [x] `body`: Java bean?
  - [x] `entity`: byte array?

whiskeysierra:
## Description

Requires #607 

## Motivation and Context

Fixes #603 

## Types of changes

## Checklist:

### DIFF
--- a/riptide-core/src/main/java/org/zalando/riptide/SerializationPlugin.java
+++ b/riptide-core/src/main/java/org/zalando/riptide/SerializationPlugin.java
@@ -1,0 +1,40 @@
+package org.zalando.riptide;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpOutputMessage;
+
+import java.io.ByteArrayOutputStream;
+
+import static org.springframework.util.CollectionUtils.toMultiValueMap;
+
+@AllArgsConstructor
+final class SerializationPlugin implements Plugin {
+
+    private final MessageWriter writer;
+
+    @Override
+    public RequestExecution aroundSerialization(final RequestExecution execution) {
+        return arguments -> {
+            final HttpHeaders headers = new HttpHeaders();
+            headers.addAll(toMultiValueMap(arguments.getHeaders()));
+            final ByteArrayOutputStream stream = new ByteArrayOutputStream();
+            final HttpOutputMessage output = new CapturingHttpOutputMessage(headers, stream);
+
+            writer.write(output, arguments);
+
+            return execution.execute(arguments
+                    .replaceHeaders(headers)
+                    .withEntity(stream.toByteArray()));
+        };
+    }
+
+    @AllArgsConstructor
+    @Getter
+    private static class CapturingHttpOutputMessage implements HttpOutputMessage {
+        private final HttpHeaders headers;
+        private final ByteArrayOutputStream body;
+    }
+
+}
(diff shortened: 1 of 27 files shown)


riptide-core/src/main/java/org/zalando/riptide/SerializationPlugin.java
```diff
@@ -0,0 +1,40 @@
...
+            final HttpHeaders headers = new HttpHeaders();
+            headers.addAll(toMultiValueMap(arguments.getHeaders()));
+            final ByteArrayOutputStream stream = new ByteArrayOutputStream();
+            final HttpOutputMessage output = new CapturingHttpOutputMessage(headers, stream);
+
+            writer.write(output, arguments);
+
+            return execution.execute(arguments
+                    .replaceHeaders(headers)
+                    .withEntity(stream.toByteArray()));
```
lukasniemeier-zalando: I wonder how often we now copy the request body. Two times or three times?

lukasniemeier-zalando:
:+1:

riptide-core/src/main/java/org/zalando/riptide/SerializationPlugin.java
whiskeysierra: This should be the first time within Riptide now. I'm thinking about using an `OutputStream` instead but it's certainly harder to use then.

whiskeysierra: Logbook would do another copy, unless we implement it natively within Riptide #604 and re-use the copy of the entity.

riptide-httpclient/src/main/java/org/zalando/riptide/httpclient/GzipHttpRequestInterceptor.java
```diff
@@ -34,7 +35,7 @@ public void process(final HttpRequest original, final HttpContext context) {
 
     private boolean hasNonEmptyBody(final HttpEntityEnclosingRequest request) {
         return Optional.ofNullable(request.getEntity())
```
lukasniemeier-zalando: just `request.getEntity() != null`?

lukasniemeier-zalando:
:+1:

riptide-httpclient/src/main/java/org/zalando/riptide/httpclient/GzipHttpRequestInterceptor.java
whiskeysierra: :man_facepalming: yes...

PR #622, opened 2019-03-07
Made caches non-shared by default

### LINKED ISSUE #621: Non-shared caches by default
## Detailed Description

Use non-shared HTTP caches by default.

## Context

Shared caches do not cache any request with an authorization header, unless the server overrides it. We never have responses that vary based on the Authorization header, therefore it's a safe default for us and makes the life of engineers easier by using the right default.

## Possible Implementation

`Caching#shared` should default to `false`.

whiskeysierra:
## Motivation and Context

Fixes #621

## Types of changes

## Checklist:


lukasniemeier-zalando:
:+1:

PR #624, opened 2019-03-08
Extracted riptide-idempotency module

### LINKED ISSUE #450: Allow for more sophisticated error handling patterns
## Detailed Description

Assuming #449 is done, we will have exactly two ways to handle erors:
1. Error responses are handled within the routing tree and maybe translated into exceptions
2. Exceptions, thrown before the response was received or during, will optionally be handled by using `CompletableFuture.exceptionally(..)`.

The idea is to give users more freedom and support when dealing with exceptions at different times of the request/response-lifecycle:

1. **Pre-response** exceptions
    This includes any kind of exception that is thrown before a response was received from the server. Those exceptions can be further split up into two different types:
   1. *Pre-request* exceptions
      Exceptions that happen exclusively on the client while the request never left the client and therefore never made it to the server in the first place. Examples include `UnknownHostException` and `NoRouteToHostException`
   2. *Post-request* exceptions
      Exceptions that may happen after the request was sent but before a response was received. Examples include `SocketTimeoutException` and `NoHttpResponseException`.
2. **Post-response** exceptions
    Exceptions that are translated from a valid HTTP response. A prominent example for this is the [`ProblemRoute`](https://github.com/zalando/riptide/tree/master/riptide-problem#usage).

General rules/assumptions:
- 1.1 can be safely retried, without any further conditions
- 1.2 should only automatically be retried when the request was idempotent, see #440 
- clients may want to override the default behavior for 1.2. by specifying on call site that retry is safe. A common example is Nakadi's POST to commit cursors which is in fact idempotent, but it's not detectable. #443 suggest something like this on client-level, but offering it on call site is the better approach.
- 2 is already fully covered since it's part of the normal response routing tree
- In the end all exceptions converge into a `CompletableFuture` and can be inspected/handled/used for fallbacks using `exceptionally(Function)`, see also [Faux Pas's `partially`](https://github.com/zalando/faux-pas#completablefutureexceptionallyfunction)

### Suggested API

```java
final Capture<Reservation> capture = Capture.empty();
return join(http.patch("/reservations/{id}", reservationId)
        .header("X-Execution-Environment", getExecutionEnvironment())
        .body(transition)
        .failure(
                on(SSLHandshakeException.class).call(this::foobar),
                anyFailure().call(retry())) // configures retry plugin for this request
        .dispatch(series(),
                on(HttpStatus.Series.SUCCESSFUL).call(Reservation.class, capture),
                on(HttpStatus.Series.SERVER_ERROR)
                        .dispatch(status(),
                                on(HttpStatus.GATEWAY_TIMEOUT).call(retry()),
                                anyStatus().call(problemHandling())),
                anySeries().call(problemHandling()))
        .thenApply(capture));
```

## Context

Doing this will allow us to separate exception handling pre- and post-receive. It will also allow clients to override exception handling on a per-call basis, effectively allowing e.g. to retry calls that otherwise wouldn't be retried.

## Possible Implementation

- Introduce `ExceptionRoutingTree` (or something along those lines)
- Introduce `ExceptionRoute` (or something along those lines)
- allow to define an exception routing tree as part of the requester
  (we most likely need to do this before the routing tree since the return value is a `CompletableFuture` already, i.e. as shown in the [API suggestion](#suggested-api))
- default exception routing tree may be just calling `CompletableFuture.completeExceptionally(Throwable)`
- rename `Plugin.beforeSend` to `Plugin.beforeNetwork`?
- general algorithm
  1. create request execution based on `AsyncClientHttpRequest.executeAsync`
  2. transform execution via `Plugin.beforeNetwork`
  3. transform execution via default exception routing tree (skip if user-supplied?)
  4. transform execution via user-supplied exception routing tree
  5. transform execution via `Plugin.beforeDispatch`
  3. transform execution via response routing tree
- Upgrade FailsafePlugin
  - implement `beforeNetwork` with algorithm as desribed in 1.1 (fix?) and 1.2 (customizable)
  - implement `beforeDispatch` with support for retry route
- Upgrade TransientFaultPlugin
  - implement `beforeNetwork` with algorithm as described in 1.1 (fix?) and 1.2 (customizable)
  - implement `beforeDispatch` with stuff like 429, Retry-After, 503, ...

## Your Environment

* Version used: 2.9.0-RC.4

### LINKED ISSUE #611: Allow negative idempotency override
## Context

> Difficult question. Having seen so many API violating the HTTP spec I would most likely (a) not make any idempotency assumptions beyond the basic assumptions implied by the HTTP method (and in this case then also assume that an additional match header doesn't violate that spec) and (b) make sure that a client has means to disble such implicit logic in the library for individual requests, in case it interacts with an API that misbehaves. It might be worth to check the proxy/cache related parts in the HTTP spec,  because what you plan to do is somehow similar to what such intermediaries have to honor. Maybe you can derive some additional hints from there.

whiskeysierra:
Closes #450
Closes #611

## Types of changes

## Checklist:


lukasniemeier-zalando:
:+1:

PR #632, opened 2019-03-14
Added support for X-RateLimit-Reset

### LINKED ISSUE #629: Support X-RateLimit-Reset for retry delays
## Detailed Description

Support `X-RateLimit-Reset` in addition to the `Retry-After` header to delay retries based on the server's suggestion.

## Context

Some servers don't use `RetryAfter` but rather `X-RateLimit-Reset` with the same semantic.

https://github.com/zalando/restful-api-guidelines/issues/379 gives some context how others are using this header.

https://opensource.zalando.com/restful-api-guidelines/#153

## Possible Implementation

- new `RateLimitResetDelayFunction`
- new `CompositeDelayFunction`  
  cycle through a collection of delay functions and use the first one that computes something
- reuse `SecondsDelayParser`
- decide whether to employ a heuristic to detect epoch seconds and introduce a new `EpochSecondsDelayParser`
- register it automatically in the Auto Configuration

whiskeysierra:
## Motivation and Context

Fixes #629

## Types of changes

## Checklist:

@jbellmann

### DIFF
--- a/riptide-failsafe/src/main/java/org/zalando/riptide/failsafe/EpochSecondsDelayParser.java
+++ b/riptide-failsafe/src/main/java/org/zalando/riptide/failsafe/EpochSecondsDelayParser.java
@@ -1,0 +1,62 @@
+package org.zalando.riptide.failsafe;
+
+import com.google.common.base.CharMatcher;
+
+import javax.annotation.Nullable;
+import java.time.Clock;
+import java.time.Duration;
+import java.time.Instant;
+
+import static java.lang.Long.parseLong;
+import static java.time.Duration.between;
+import static java.time.Instant.now;
+
+final class EpochSecondsDelayParser implements DelayParser {
+
+    private final CharMatcher digit = CharMatcher.inRange('0', '9').precomputed();
+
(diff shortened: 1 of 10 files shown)


riptide-failsafe/src/main/java/org/zalando/riptide/failsafe/EpochSecondsDelayParser.java
```diff
@@ -0,0 +1,50 @@
...
+    @Override
+    public Duration parse(final String value) {
+        return until(parseDate(value));
+    }
+
+    @Nullable
+    private Instant parseDate(final String value) {
+        if (isInteger(value)) {
+            final long epochSecond = parseLong(value);
+            if (epochSecond > 1_000) {
```
lukasniemeier-zalando: What about we (dynamically?) compare with `now() - '1 day'` and decide based on this whether it's seconds vs epoch timestamp? This way we'd allow servers to configure delays a la '1 hour'.

lukasniemeier-zalando:
:+1:

PR #639, opened 2019-03-29
Added compatibility module

### LINKED ISSUE #631: Compatibility Module
## Detailed Description

Introduce a dedicated module that contains adapters/implementations for users familiar with Spring's default means of interaction (Async/RestTemplate, WebServiceTemplate?) to ease their transition to Riptide.

## Context

Some users prefer the simplified syntax from RestTemplates over Riptides rich but (perceived as) verbose API. In order to attract those users we may choose to offer better first-class support.

The current support of `PluginInterceptor` is insufficient. The semantic mismatch between Plugins and interceptors is too big to feel confident that it will work in the future. Instead of trying to squeeze plugins into RestTemplates we should rather think of another layer just put on top of Riptide allowing to make use of all the features consistently and not just some of them (e.g. URI resolution).

## Possible Implementation

- [x] Drop `PluginInterceptor` from `riptide-core`
- [x] New module `riptide-compatibility` (or `riptide-compat`?)
- [x] `HttpOperations` adapter which implements `RestOperations` by delegating to `Http` (+ `Capture`)
- [x] `AsyncHttpOperations` adapter which implements `AsyncRestOperations` by delegating to `Http`
- [x] Drop registration of `RestTemplate` and `AsyncRestTemplate` in auto configuration
- [ ] Register both adapters for each client (configurable?)
- [ ] Adjust `ManualConfiguration`

whiskeysierra:
## Motivation and Context

Fixes #631 

## Types of changes

## Checklist:

- [X] My change requires a change to the documentation.


whiskeysierra:
Possible readme banner https://pixabay.com/photos/miniature-car-model-toy-automobile-1802333/

lukasniemeier-zalando:
:+1:

PR #678, opened 2019-05-31
Added NIO support

whiskeysierra:
## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ Feel free to compare this e.g. to [Feign](https://github.com/Netflix/feign#basic
   - retries and circuit breaker via [Failsafe integration](riptide-failsafe)
   - backup requests via [riptide-backup](riptide-backup)
   - [timeouts](riptide-timeout)
+- non-blocking IO (optional)
 - encourages the use of
   - fallbacks
   - content negotiation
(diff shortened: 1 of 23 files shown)


whiskeysierra:
@lukasniemeier-zalando @tkrop :+1::question:

tkrop:
Looks good, but I will have a second look before I give my :+1:.

tkrop:
I absolutely like this, but I'm missing the documentation :).

whiskeysierra:
@tkrop I added a section to the README. Can you give me a 👍?

tkrop:
With this explanation you have confused me - I can't figure out how to apply the feature to its core use case. I'm missing the motivation and explanation how the Async IO helps to implement processor intensive execution chains based on the common thread pool.

whiskeysierra:
> I'm missing the motivation and explanation how the Async IO helps to implement processor intensive execution chains based on the common thread pool.

I'm just trying to offer it for those interested in it. I don't believe this library is the right place to motivate and advocate for non-blocking IO.

> I can't figure out how to apply the feature to its core use case.

You just need to configure an `AsyncClientHttpRequestFactory` implementation of your choice. What would you like to see?

lukasniemeier-zalando:
:+1:

tkrop:
@whiskeysierra Sorry for not responding faster.

I think, I would have liked to see a two sentence explanation how to make use of a common executor service to circumvent the [How to set an ideal thread pool size](https://jobs.zalando.com/tech/blog/how-to-set-an-ideal-thread-pool-size/) issue you referred to directly above.

In my opinion we should explain, that the core idea of the non-blocking IO support is a clear separation of *wait time* and *service time* handling to simplify thread pool sizing. This is done by combining the non-blocking IO request factory with a single(ton) processor-number-sized executor services for optimal proportion between memory consumption and thread starvation.

Besides, I find the table a bit confusing, because it is telling only have the story neglecting the `capture` extension. Doing so it only shows the trivial facts. In addition, it is not showing the most favorable usage option outlined above.

whiskeysierra:
For non blocking IO you don't provide any executor to riptide at all. That's completely up to the implementation of the factory. The API doesn't actually allow you to specify an executor. The table is misleading there, agreed.

PR #689, opened 2019-06-13
Cleanup of IdempotencyPlugin declaration

fatroom:
Cleanup of IdempotencyPlugin declaration

## Motivation and Context
As soon as I got riptide-idempotence in my class path, I can not create simple http client without specifying plugins any more. Default plugin resolver see plugin declaration in META-INF.services but fails to find actual class, which leads to ```ServiceConfigurationException: org.zalando.riptide.Plugin: Provider org.zalando.riptide.idempotency.IdempotencyPlugin not found.```

## Types of changes

## Checklist:

### DIFF
--- a/riptide-idempotency/src/main/java/org/zalando/riptide/idempotency/IdempotencyPlugin.java
+++ b/riptide-idempotency/src/main/java/org/zalando/riptide/idempotency/IdempotencyPlugin.java
@@ -1,0 +1,6 @@
+package org.zalando.riptide.idempotency;
+
+import org.zalando.riptide.Plugin;
+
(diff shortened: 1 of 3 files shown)


whiskeysierra:
Good catch! The proper fix would be to just remove this file: https://github.com/zalando/riptide/blob/master/riptide-idempotency/src/main/resources/META-INF/services/org.zalando.riptide.Plugin

fatroom:
Adjusted accordingly :)

whiskeysierra:
:+1:

whiskeysierra:
Thanks for the contribution! I'm releaseing 3.0.0-RC.2 right now :package:

PR #694, opened 2019-06-17
Inject tracer by bean reference instead of name

fatroom:
## Description
Inject tracer by bean reference instead of name

## Motivation and Context
Currently tracer bean is injected by name. This leads to incompatibility issues with tracing-lightstep-spring-boot-starter > 0.1.5
Instead of relying on bean name for plugin registration, I've changed to loading by type.

## Types of changes

## Checklist:

### DIFF
--- a/riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/DefaultRiptideRegistrar.java
+++ b/riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/DefaultRiptideRegistrar.java
@@ -5,6 +5,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.binder.jvm.ExecutorServiceMetrics;
+import io.opentracing.Tracer;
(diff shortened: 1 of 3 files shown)


fatroom:
Still I'm not really sure how to test the context that will fail to load on missing bean declaration. Any ideas?

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/DefaultRiptideRegistrar.java
```diff
@@ -168,10 +170,11 @@ private String registerExecutor(final String id, final Client client) {
         }
 
         if (client.getTracing().getEnabled()) {
+            BeanReference tracerRef = registry.findRef(null, Tracer.class).orElseThrow(() -> new NoSuchBeanDefinitionException(Tracer.class));
```
whiskeysierra: `findRef` still searches by name. It's just using the given class to construct a name. There is no easy way to autowire by type when dealing with a bean definition registry. It's one of the downsides of how the auto configuration is implemented right now.

fatroom: thanks for pointing out :/

PR #695, opened 2019-06-17
Added resolving beans by type

fatroom:
## Description
Inject tracer by bean reference instead of name

## Motivation and Context
Currently tracer bean is injected by name. This leads to incompatibility issues with tracing-lightstep-spring-boot-starter > 0.1.5
Instead of relying on bean name for plugin registration, I've changed to loading by type.

## Types of changes

## Checklist:


riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/RiptidePostProcessor.java
```diff
@@ -34,13 +35,13 @@ public void setEnvironment(final Environment environment) {
 
     @Override
     public void postProcessBeanDefinitionRegistry(final BeanDefinitionRegistry registry) {
-        final RiptideRegistrar registrar = registrarFactory.apply(new Registry(registry), Defaulting.withDefaults(properties));
-        registrar.register();
+        this.registry = registry;
     }
 
     @Override
     public void postProcessBeanFactory(final ConfigurableListableBeanFactory beanFactory) throws BeansException {
```
whiskeysierra: Interesting that this seems to work since the javadoc for `postProcessBeanFactory` actually state:

> All bean definitions will have been loaded, but no beans will have been instantiated yet. This allows for overriding or adding properties even to eager-initializing beans.

To me that reads like the only thing left to do is modifying properties, but not register new bean definitions.

fatroom: At least reference doesn't state anything about prohibition of beans creation at this point. Spring it self using casting from BeanDefinitionRegistry to ConfigurableListableBeanFactory in postProcessBeanDefinitionRegistry, but not sure if it worth to go this way. Another idea would be perform class path scanning, but that's seems to be overkill

whiskeysierra: Using bean factory was my initial idea but it didn't offer any ways to register new beans (definitions). It didn't occur to me that you can work with both in conjunction.

whiskeysierra:
I believe we have a couple of other places where this could be useful. Basically whenever we inject by name right now. Logbook comes to mind, e.g.

fatroom:
Another way that came into my mind: register factories in postProcessBeanDefinitionRegistry and afterwards configure them in postProcessBeanFactory. This seems to be the most API-friendly way of solving the problem, but will require a lot(?) of changes in initialisation

whiskeysierra:
> I believe we have a couple of other places where this could be useful. Basically whenever we inject by name right now. Logbook comes to mind, e.g.

And `MeterRegistry`.

> Another way that came into my mind: register factories in postProcessBeanDefinitionRegistry and afterwards configure them in postProcessBeanFactory. This seems to be the most API-friendly way of solving the problem, but will require a lot(?) of changes in initialisation

We have two kinds of configuration. Some of them completely disable a bean and others really just configure it. So we would need to spread this across both.

fatroom:
I would propose following:
- introduce properties for *PluginFactories
- leave creation of the factories for the postProcessBeanDefinitionRegistry so we got all the beans registered during appropriate phases
- setup the properties from the postProcessBeanFactory so we got type resolving available for us.

If this sounds fine for you I'll start working on implementation.

> We have two kinds of configuration. Some of them completely disable a bean and others really just configure it. So we would need to spread this across both.

I don't get this point. What do you mean by two kinds of configuration? Any sample in the code?

whiskeysierra:
> I don't get this point. What do you mean by two kinds of configuration? Any sample in the code?

```yaml
riptide.clients.example:
  tracing:
    enabled: true
    tags:
      peer.service: example-service
```

The `enabled` property dictates that the bean should be registered (or not, if set to `false`) while the rest (e.g. `tags`) is meant to configure the bean (in this case `OpenTracingPlugin`).

So the configuration affects both the registration and wiring of beans as well as the configuration of the properties of said beans.

whiskeysierra:
> If this sounds fine for you I'll start working on implementation.

Looking forward to some code to discuss this further! :tada:

fatroom:
While implementing a lot of factories, I just recognised that we don't actually require that. So there's a lot more simplified approach. I haven't changed yet the metricsRegistry yet, since it leads to broken tests, and I want to figure out reasons for this behaviour.

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/DefaultRiptideConfigurer.java
```diff
@@ -0,0 +1,67 @@
...
+        } catch (NoSuchBeanDefinitionException e) {
+            return Optional.empty();
+        }
+    }
+
+    private void replaceConstructorArgumentWithBean(BeanDefinition bd, String arg, BeanDefinition ref) {
+        bd.getConstructorArgumentValues()
+          .getIndexedArgumentValues()
+          .values().stream()
+          .filter(holder -> arg.equals(holder.getValue()))
```
whiskeysierra: That feels risky.

fatroom: yep. There's two options: 

1. rely on position index. 
1. Use a cryptic constant for default value similar to https://github.com/spring-projects/spring-framework/blob/3a0f309e2c9fdbbf7fb2d348be861528177f8555/spring-messaging/src/main/java/org/springframework/messaging/handler/annotation/ValueConstants.java#L34

fatroom:
So with moving away from fixed name strategy, appropriate bean resolving will happen based on `primary` annotation if available, or the last one in the found beans. 
Still I believe this approach is better than relying on the "bean name never change" strategy, as we know that this is not the case.

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/DefaultRiptideConfigurer.java
```diff
@@ -0,0 +1,99 @@
...
+
+            findBeanDefinition(id, RetryListener.class)
+                    .ifPresent(bd -> replaceConstructorArgumentWithBean(bd, METER_REGISTRY_REF, meterRegistryRef));
+
+            findBeanDefinition(id, CircuitBreakerListener.class)
+                    .ifPresent(bd -> replaceConstructorArgumentWithBean(bd, METER_REGISTRY_REF, meterRegistryRef));
+        }
+    }
+
+    private BeanDefinition getBeanRef(Class type) {
```
whiskeysierra:
```java
return Stream.of(beanFactory.getBeanNamesForType(type))
        .map(beanFactory::getBeanDefinition)
        .sorted(comparing(BeanDefinition::isPrimary).reversed())
        .findFirst().orElse(null);
```

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/DefaultRiptideConfigurer.java
```diff
@@ -0,0 +1,99 @@
...
+        return definition;
+    }
+
+    private Optional<BeanDefinition> findBeanDefinition(final String id, final Class<?> type) {
+        return findBeanDefinition(Name.name(id, type));
+    }
+
+    private Optional<BeanDefinition> findBeanDefinition(final Name name) {
+        try {
+            return Optional.of(beanFactory.getBeanDefinition(name.toNormalizedString()));
```
whiskeysierra: Shouldn't this search for all possible spellings (`getAlternatives`)?

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/DefaultRiptideConfigurer.java
```diff
@@ -0,0 +1,99 @@
...
+
+        if (client.getLogging().getEnabled()) {
+            final BeanDefinition logbookRef = getBeanRef(Logbook.class);
+
+            findBeanDefinition(id, LogbookPlugin.class)
+                    .ifPresent(bd -> replaceConstructorArgumentWithBean(bd, LOGBOOK_REF, logbookRef));
+        }
+
+        if (client.getMetrics().getEnabled()) {
+            final BeanDefinition meterRegistryRef = getBeanRef(MeterRegistry.class);
```
whiskeysierra: This may find nothing and return `null` which means we replace the argument with `null` and then startup with a broken setup. Shouldn't this be throwing if e.g. metrics is enabled but no `MeterRegistry` was found.

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/DefaultRiptideConfigurer.java
fatroom: Don't think so. 
Alternative names can appear from plugins that we're overwritten by user, in such case it would be a bit unfair to mess with user configuration. And if bean was registered by the registar - the normalisedname is used.

whiskeysierra: Right

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/RiptidePostProcessor.java
```diff
@@ -40,7 +40,8 @@ public void postProcessBeanDefinitionRegistry(final BeanDefinitionRegistry regis
...
-        // nothing to do
+        final DefaultRiptideConfigurer configurer = new DefaultRiptideConfigurer(beanFactory, Defaulting.withDefaults(properties));
```
whiskeysierra: `Defaulting.withDefaults(properties)` can be done once in `setEnvironment`

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/ValueConstants.java
```diff
@@ -0,0 +1,14 @@
+package org.zalando.riptide.autoconfigure;
+
+/**
+ * Constants refering to beans that need to be looked up by types during Riptide clients constructions.
+ * <p>This is an artificial arrangement of 16 unicode characters, with its sole purpose
+ * being to never match user-declared values.
+ */
+class ValueConstants {
```
whiskeysierra: This deserves a Google Gag remark annotation, e.g. `@Magic`.

fatroom: Added magic, but it seems to me that Hack is more applicable here

whiskeysierra:
:+1:

fatroom:
@whiskeysierra while I was describing the changes in migration guide, something struck in my mind. Afaik if multiple beans of the same type present in spring context, but no primary explicitly defined I think spring will fail the context init. But last time I saw this behaviour was somewhere during initial release of spring 4, and not sure if it's current one. Should we follow this behaviour as well here, instead of injection of the first random bean?

whiskeysierra:
I believe Spring tries to fallback to injecting by name. If that fails is usually complains about non-unique beans. Ideally we would hand over all of this work to Spring.

fatroom:
Switched to default spring behaviour: try to find primary bean, if not found use resolve by name.

whiskeysierra:
You're missing some code coverage there.

whiskeysierra:
:+1:

PR #698, opened 2019-06-18
Switched to immutable builder

whiskeysierra:
## Types of changes

## Checklist:


fatroom:
What the reasoning for conversion to immutable builder?

whiskeysierra:
Consistency with `Http` itself and generally I prefer immutable builders because you can freely re-use them without risk. Let's say someone configures `Http` instances programmatically then you could start with a single builder, configure the common basics and "branch" off from there, even allowing to override stuff without affecting any other client.

fatroom:
👍

PR #702, opened 2019-06-23
Added proper support for stack and message span tags

whiskeysierra:
## Description

Added new span decorators for `stack` and `message` log fields:
https://github.com/opentracing/specification/blob/master/semantic_conventions.md

## Motivation and Context

Consistency with internal OpenTracing semantic conventions. @pc-alves

## Types of changes

## Checklist:

### DIFF
--- a/riptide-opentracing/src/main/java/org/zalando/riptide/opentracing/OpenTracingPlugin.java
+++ b/riptide-opentracing/src/main/java/org/zalando/riptide/opentracing/OpenTracingPlugin.java
@@ -17,6 +17,7 @@ import org.zalando.riptide.RequestExecution;
 import org.zalando.riptide.opentracing.span.CallSiteSpanDecorator;
 import org.zalando.riptide.opentracing.span.ComponentSpanDecorator;
 import org.zalando.riptide.opentracing.span.ErrorSpanDecorator;
+import org.zalando.riptide.opentracing.span.ErrorStackSpanDecorator;
 import org.zalando.riptide.opentracing.span.HttpMethodSpanDecorator;
 import org.zalando.riptide.opentracing.span.HttpPathSpanDecorator;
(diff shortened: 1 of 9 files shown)


riptide-opentracing/src/main/java/org/zalando/riptide/opentracing/OpenTracingPlugin.java
```diff
@@ -83,6 +84,7 @@ public OpenTracingPlugin(final Tracer tracer) {
                         new CallSiteSpanDecorator(),
                         new ComponentSpanDecorator(),
                         new ErrorSpanDecorator(),
+                        new ErrorStackSpanDecorator(),
```
pc-alves: Just to confirm: Are you purposely leaving out the `ErrorMessageSpanDecorator`?

whiskeysierra: Yes, but I'm open for discussions.

https://github.com/zalando/riptide/pull/702/files#diff-72802c1148338662d60bb7c4ee341fe0R56 this will call `Throwable#toString` which already includes `getMessage`. So in other words the message is already part of the stack (as is normal for Java applications). But I wouldn't mind adding it, just to have it a) as a separate log field and b) be consistent with other languages that don't have the same idea of a stack trace.

We could also change the rendering of the stack trace and remove messages from there.

pc-alves: > this will call `Throwable#toString` which already includes `getMessage`

This being the case, I agree that it makes sense to leave out the `ErrorMessageSpanDecorator`.
Library users still get a log field with the exception message, and they still have the `message` log field to add what ever additional information they wish.

fatroom:
👍

PR #707, opened 2019-07-04
Removed onStart decorator method

whiskeysierra:
It only worked with the NewSpanLifecyclePolicy and was incompatible with the other policies

## Types of changes

## Checklist:

### DIFF
--- a/riptide-opentracing/src/main/java/org/zalando/riptide/opentracing/CompositeLifecyclePolicy.java
+++ b/riptide-opentracing/src/main/java/org/zalando/riptide/opentracing/CompositeLifecyclePolicy.java
@@ -4,10 +4,9 @@ import io.opentracing.Span;
 import io.opentracing.Tracer;
 import lombok.AllArgsConstructor;
 import org.zalando.riptide.RequestArguments;
-import org.zalando.riptide.opentracing.span.SpanDecorator;
 
-import javax.annotation.Nullable;
 import java.util.Collection;
+import java.util.Optional;
 import java.util.function.Consumer;
 
 @AllArgsConstructor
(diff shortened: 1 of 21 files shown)


riptide-opentracing/src/main/java/org/zalando/riptide/opentracing/CompositeLifecyclePolicy.java
```diff
@@ -4,29 +4,27 @@
...
     @Override
-    public Span start(final Tracer tracer, final RequestArguments arguments, final SpanDecorator decorator) {
+    public Optional<Span> start(final Tracer tracer, final RequestArguments arguments) {
         for (final LifecyclePolicy policy : policies) {
-            @Nullable final Span span = policy.start(tracer, arguments, decorator);
+            final Optional<Span> span = policy.start(tracer, arguments);
 
-            if (span != null) {
-                return new FinishingSpan(span, policy::finish);
+            if (span.isPresent()) {
```
fatroom: seems to me that span.map(s -> new FinishingSpan(s, policy::finish)) would be shorter here

whiskeysierra: I'd still need to use `if (x.isPresent()) {return x.get();}` afterwards since I need to return from the outer method which I can't do within a lambda.

fatroom: ah, missed that return will be executed only in case of presence

riptide-opentracing/src/main/java/org/zalando/riptide/opentracing/OpenTracingPlugin.java
```diff
@@ -78,7 +78,11 @@
     private final SpanDecorator decorator;
 
     public OpenTracingPlugin(final Tracer tracer) {
-        this(tracer, LifecyclePolicy.composite(new ExplicitSpanLifecyclePolicy(), new NewSpanLifecyclePolicy()), new DefaultActivationPolicy(),
+        this(tracer,
+                LifecyclePolicy.composite(
+                        new ExplicitSpanLifecyclePolicy(),
+                        new NewSpanLifecyclePolicy()),
+                new DefaultActivationPolicy(),
```
fatroom: formatting is off here

whiskeysierra: No, it's correct. `DefaultActivationPolicy` is a separate parameter, next to `LifecyclePolicy.composite` and `SpanDecorator.composite`.

fatroom:
👍

Issue #714, opened 2019-07-16
Unexpected dependency on httpclient-cache


lukasniemeier-zalando:
## Description

Using the spring-boot-starter `httpclient-cache` seems to be a mandatory dependency, despite no `caching` is enabled.

## Expected Behavior

If no `caching` is configured, `httpclient-cache` should not be a mandatory depdendency.

## Actual Behavior

In `DefaultRiptideRegistrar` the class `HttpCacheStorage` is always loaded, failing if `httpclient-cache` is not on the classpath.

## Possible Fix

Only try to load the class if dependency is available.

## Steps to Reproduce
1. Add the spring-boot-starter to your app
2. Configure a client (without `caching`)
3. See `java.lang.NoClassDefFoundError: org/apache/http/client/cache/HttpCacheStorage`

## Your Environment

Running version `3.0.0-RC.2`.

fatroom:
Actually the problem lying not in the DefaultRiptideRegistrar, but in the HttpClientFactory which unconditionally references HttpCacheStorage. 

So either we need to re-write this Factory somehow, or make httpclient-cache hard dependency. 
What would be the preferred way?

whiskeysierra:
> What would be the preferred way?

I'd like to have it as a real optional dependency.

lukasniemeier-zalando:
I somehow remember that we had something like this in the past already, we used `Object` instead of the actual (optional) type before casting it in the branch it eventually got used.

whiskeysierra:
See #716

lukasniemeier-zalando:
I am still seeing the unexpected dependency with Riptide `3.0.0-RC.3`

```
java.lang.IllegalStateException: Failed to load ApplicationContext
	at org.springframework.test.context.cache.DefaultCacheAwareContextLoaderDelegate.loadContext(DefaultCacheAwareContextLoaderDelegate.java:125)
	at org.springframework.test.context.support.DefaultTestContext.getApplicationContext(DefaultTestContext.java:108)
	at org.springframework.test.context.support.DependencyInjectionTestExecutionListener.injectDependencies(DependencyInjectionTestExecutionListener.java:118)
	at org.springframework.test.context.support.DependencyInjectionTestExecutionListener.prepareTestInstance(DependencyInjectionTestExecutionListener.java:83)
[...]

PR #716, opened 2019-07-17
Made reference to httpclient-cache optional

fatroom:
## Description
Proposal for fix of #714

## Motivation and Context
At the moment httpclient-cache is hard dependency, even if it's not declared explicitly in pom as such.
This fix intent to make riptide independent from this dependency if caching is not enabled.

## Types of changes

## Checklist:

### DIFF
--- a/riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/DefaultRiptideRegistrar.java
+++ b/riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/DefaultRiptideRegistrar.java
@@ -600,7 +600,7 @@ final class DefaultRiptideRegistrar implements RiptideRegistrar {
                     .addConstructorArgValue(configureFirstRequestInterceptors(id, client))
                     .addConstructorArgReference(connectionManager)
                     .addConstructorArgValue(registry.findRef(id, HttpClientCustomizer.class).orElse(null))
-                    .addConstructorArgValue(registry.findRef(id, HttpCacheStorage.class).orElse(null))
+                    .addConstructorArgValue(findCacheStorageReference(id, client).orElse(null))
                     .setDestroyMethodName("close");
         });
     }
@@ -614,6 +614,14 @@ final class DefaultRiptideRegistrar implements RiptideRegistrar {
         }
 
         return interceptors;
+    }
+
+    private Optional<BeanReference> findCacheStorageReference(final String id, final Client client) {
+        if (client.getCaching().getEnabled()) {
+            return registry.findRef(id, HttpCacheStorage.class);
+        } else {
+            return Optional.empty();
+        }
     }
 
 }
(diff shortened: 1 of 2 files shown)


fatroom:
I have no idea on how to write tests for this change (locally tested with separate project without http client in class path).
Any advices appreciated.

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/HttpClientFactory.java
```diff
@@ -97,8 +97,8 @@ public static CloseableHttpClient createHttpClient(final Client client,
         return builder.build();
     }
 
-    private static CachingHttpClientBuilder configureCaching(final Caching caching,
-            @Nullable final HttpCacheStorage cacheStorage) {
+    private static HttpClientBuilder configureCaching(final Caching caching,
+            @Nullable final Object cacheStorage) {
```
whiskeysierra: Can be already typed here and the cast can be pushed upwards to the public method.

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/DefaultRiptideRegistrar.java
```diff
@@ -616,4 +616,12 @@ private String registerHttpClient(final String id, final Client client) {
         return interceptors;
     }
 
+    private Optional<BeanReference> findCacheStorageReference(final String id, final Client client) {
```
whiskeysierra: :+1:

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/HttpClientFactory.java
fatroom: Not really. If I leave HttpCacheStorage here I'm receiving:
```
Failed to introspect Class [org.zalando.riptide.autoconfigure.HttpClientFactory] from ClassLoader [jdk.internal.loader.ClassLoaders$AppClassLoader@2c13da15]
	at org.springframework.util.ReflectionUtils.getDeclaredMethods(ReflectionUtils.java:507) ~[spring-core-5.1.8.RELEASE.jar:5.1.8.RELEASE]
```

Seems that even if the method is private somehow spring still tries to evaluate it.

fatroom:
I found a suggestion on how to exclude dependency for test: https://stackoverflow.com/questions/16019550/exclude-maven-dependency-for-one-specific-unit-test

but there's on downside: IDEA will not recognise such type of configuration https://youtrack.jetbrains.com/issue/IDEA-122783

Should we still go this way? Or just merge as it is?

whiskeysierra:
> but there's on downside: IDEA will not recognise such type of configuration https://youtrack.jetbrains.com/issue/IDEA-122783

How does that manifest itself?

fatroom:
> How does that manifest itself?
Sorry, don't got your question

whiskeysierra:
What do I see in idea. Or what wouldn't work?

fatroom:
> What do I see in idea. Or what wouldn't work?

The tests will give different results in IDEA and maven. While maven test with different surefire configs will catch the problem if runs against current master, if you run the same test from idea - it will be green. I'm a bit sceptical on idea of introducing tests that will behave differently based on execution environments.

whiskeysierra:
> While maven test with different surefire configs will catch the problem if runs against current master, if you run the same test from idea

That's OK. Maven is the reference and used by Travis. So we are safe against accidents.

> I'm a bit sceptical on idea of introducing tests that will behave differently based on execution environments.

In most libraries/modules I'd agree. But especially in our `*-autoconfigure` modules the only task that we have is to integrate different modules and keep the *à la carte* offering from using modules individually. I'm OK with that.

whiskeysierra:
:+1:

PR #722, opened 2019-07-27
Added SPI support for SpanDecorators

whiskeysierra:
## Description

By default span decorators are also picked up via the service loader.

## Motivation and Context

Users that prefer to register custom `SpanDecorators` using the service provider interface (ServiceLoader) now can do that, rather than registering them by hand.

## Types of changes

## Checklist:


whiskeysierra:
Since it's not a named, individual class it can't be registered manually. Not sure if that would be useful.

fatroom:
👍

PR #757, opened 2019-08-29
Added support for more http-based span tags

whiskeysierra:
## Description

Added span decorators for more tags.

## Motivation and Context

Some http headers are particularly useful and translating them into tags allows to use them within OpenTracing directly.

## Types of changes

## Checklist:

### DIFF
--- a/riptide-opentracing/README.md
+++ b/riptide-opentracing/README.md
@@ -63,29 +63,43 @@ By default a new span will be started for each request and it will be activated.
 
 The following tags/logs are supported out of the box:
 
-| Tag/Log Field        | Decorator                      | Example                           |
-|----------------------|--------------------------------|-----------------------------------|
-| `component`          | `ComponentSpanDecorator`       | `Riptide`                         |
-| `span.kind`          | `SpanKindSpanDecorator`        | `client`                          |
-| `peer.address`       | `PeerSpanDecorator`            | `www.github.com:80`               |
-| `peer.hostname`      | `PeerSpanDecorator`            | `www.github.com`                  |
-| `peer.port`          | `PeerSpanDecorator`            | `80`                              |
-| `http.method`        | `HttpMethodSpanDecorator`      | `GET`                             |
-| `http.url`           | `HttpUrlSpanDecorator`         | `https://www.github.com/users/me` |
-| `http.path`          | `HttpPathSpanDecorator`        | `/users/{user_id}`                |
-| `http.status_code`   | `HttpStatusCodeSpanDecorator`  | `200`                             |
-| `error`              | `ErrorSpanDecorator`           | `false`                           |
-| `error.kind` (log)   | `ErrorSpanDecorator`           | `SocketTimeoutException`          |
-| `error.object` (log) | `ErrorSpanDecorator`           | (exception instance)              |
-| `message` (log)      | `ErrorMessageSpanDecorator`    | `Connection timed out`            |
-| `stack` (log)        | `ErrorStackSpanDecorator`      | `SocketTimeoutException at [...]` |
-| `retry`              | `RetrySpanDecorator`           | `true`                            |
-| `retry_number` (log) | `RetrySpanDecorator`           | `3`                               |
-| `*`                  | `CallSiteSpanDecorator`        | `admin=true`                      |
-| `*`                  | `StaticTagSpanDecorator`       | `aws.region=eu-central-1`         |
-| `*`                  | `UriVariablesTagSpanDecorator` | `user_id=me`                      |
+| Tag/Log Field                | Decorator                          | Example                           |
+|------------------------------|------------------------------------|-----------------------------------|
+|                              | `CallSiteSpanDecorator`            | `admin=true`                      |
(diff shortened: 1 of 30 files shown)


riptide-opentracing/src/main/java/org/zalando/riptide/opentracing/span/HttpRateLimitSpanDecorator.java
```diff
@@ -0,0 +1,45 @@
...
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
+import static org.zalando.riptide.opentracing.span.HttpHeaderSpanDecorator.mapping;
+
+/**
+ * @see <a href="https://ioggstream.github.io/draft-polli-ratelimit-headers/draft-polli-ratelimit-headers.html">Rate-Limit Header Fields for HTTP</a>
+ */
+@API(status = EXPERIMENTAL)
+public final class HttpRateLimitSpanDecorator extends DelegateSpanDecorator {
+
+    private static final Tag<String> LIMIT =
```
whiskeysierra: `http.rate_limit_limit`?
`http.rate_limit.limit`?

fatroom: second seems more appropriate as 'limit' will be part of 'rate_limit'

fatroom: I'm not sure to which level this will be used, but is there possibility to switch to `rate_limit.remaining` and so on?

whiskeysierra: I started an internal discussion about the proper name for this:
https://github.bus.zalan.do/SRE/opentracing/issues/43

riptide-opentracing/src/main/java/org/zalando/riptide/opentracing/span/SpanOperator.java
```diff
@@ -0,0 +1,10 @@
+package org.zalando.riptide.opentracing.span;
+
+import io.opentracing.Span;
+
+import java.util.List;
+import java.util.Map;
+
+interface SpanOperator {
+    void apply(Span span, Map<String, List<String>> headers);
```
fatroom: not the best naming example here.

How `SpanOperator` relates to `headers`? Seems to me like abstraction leak.

riptide-opentracing/README.md
```diff
@@ -63,29 +63,41 @@ By default a new span will be started for each request and it will be activated.
...
+| `http.warning`               | `HttpWarningSpanDecorator`         | `110 - "Response is Stale"`       |
+| `peer.address`               | `PeerSpanDecorator`                | `www.github.com:80`               |
+| `peer.hostname`              | `PeerSpanDecorator`                | `www.github.com`                  |
+| `peer.port`                  | `PeerSpanDecorator`                | `80`                              |
+| `rate_limit.limit` (log)     | `RateLimitSpanDecorator`           | `100`                             |
+| `rate_limit.remaining` (log) | `RateLimitSpanDecorator`           | `99`                              |
+| `rate_limit.reset` (log)     | `RateLimitSpanDecorator`           | `3600`                            |
+| `retry`                      | `RetrySpanDecorator`*              | `true`                            |
+| `retry_number` (log)         | `RetrySpanDecorator`*              | `3`                               |
+|                              | `ServiceLoaderSpanDecorator`       |                                   |
```
fatroom: RetrySpanDecorator marked with * so reader can expect some explanation after the table, but this one doesn't marked with anything and got special behaviour and representation in this table. A bit confusing, unless you'll read the whole table and find out the special behaviour description

riptide-opentracing/src/main/java/org/zalando/riptide/opentracing/span/SpanOperator.java
whiskeysierra: Totally agree. Struggled with a good design there. That includes names. Since all of this is package private I don't care that much right now, but if you have a good suggestion, I'm all for it.

riptide-opentracing/README.md
whiskeysierra: Any idea how to improve that? Footnote style with superscript numbers?

fatroom: Footnote with numbers or star markers will make this much better

riptide-opentracing/src/main/java/org/zalando/riptide/opentracing/span/SpanOperator.java
fatroom: `HttpSpanOperator` as name will provide at least context for the headers parameters here.

whiskeysierra:
@fatroom I'd like to merge this now and let the discussions about rate limit tags/logs continue on its own speed. If needed I'll change or extend it here later. :+1: :question:

fatroom:
👍

PR #758, opened 2019-08-29
Reduced minimum spring version requirement to 4.1.6

whiskeysierra:
## Description

(does not apply to spring boot starter and auto configuration)

## Motivation and Context

Allow usage of core Riptide modules in older Spring versions.

## Types of changes

## Checklist:

### DIFF
--- a/riptide-chaos/src/main/java/org/zalando/riptide/chaos/EmptyInputStream.java
+++ b/riptide-chaos/src/main/java/org/zalando/riptide/chaos/EmptyInputStream.java
@@ -1,0 +1,19 @@
+package org.zalando.riptide.chaos;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+final class EmptyInputStream extends InputStream {
+
+    static final InputStream EMPTY = new EmptyInputStream();
+
+    private EmptyInputStream() {
+
+    }
+
+    @Override
+    public int read() throws IOException {
+        return -1;
+    }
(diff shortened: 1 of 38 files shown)


riptide-chaos/src/main/java/org/zalando/riptide/chaos/EmptyInputStream.java
```diff
@@ -0,0 +1,19 @@
+package org.zalando.riptide.chaos;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+final class EmptyInputStream extends InputStream {
```
lukasniemeier-zalando: this incarnation of this class is untested :stuck_out_tongue_closed_eyes:

lukasniemeier-zalando:
:+1:

riptide-chaos/src/main/java/org/zalando/riptide/chaos/EmptyInputStream.java
whiskeysierra: How do you know?

lukasniemeier-zalando: I only see a test for the one of `riptide-httpclient`... but I also see the similarities between the implementations.

whiskeysierra:
It's covered implicitly by other unit tests

PR #741, opened 2019-08-12
Avoid dependency to CachingHttpClientBuilder

fatroom:
## Description
Second try to address issue #714 

## Motivation and Context
Seems even if the #716 was enough to trick IDEA into loading riptide successfully, but that fix was not complete in case running the application from the maven. I've added simple test to verify behaviour. This requires the new surefire configuration to manipulate class path around for test execution. 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.

PR #761, opened 2019-09-02
Revert "Bump spring.version from 4.1.6.RELEASE to 5.1.9.RELEASE"

whiskeysierra:
This reverts commit df88756495ccf3796c17c7685b2df903a3b41649.

## Motivation and Context

#758 was meant to stay.

## Types of changes

## Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,7 +74,7 @@
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
         <slf4j.version>1.7.28</slf4j.version>
-        <spring.version>5.1.9.RELEASE</spring.version>
+        <spring.version>4.1.6.RELEASE</spring.version>
         <spring-boot.version>2.1.7.RELEASE</spring-boot.version>
         <jackson.version>2.10.0.pr2</jackson.version>
         <problem.version>0.23.0</problem.version>


whiskeysierra:
I can't find an enforcer rule to require a certain dependency. And by design I want my code to be binary compatible to Spring 4 and Spring 5, so an upgrade will always work.

I could tell dependabot not to update to 5.x, but that's still risky.

fatroom:
What about setting separate set of test dependencies like I did with the #741 ? So we can have to test runs for spring 4 and another for spring 5?

whiskeysierra:
I could configure dependabot with a file.

https://dependabot.com/docs/config-file/

whiskeysierra:
> What about setting separate set of test dependencies like I did with the #741 ? So we can have to test runs for spring 4 and another for spring 5?

I used profiles for that but since my default is spring4 it can silently be upgraded and both are the same. I could have a separate profile or test execution which uses a hard coded spring4 version. That might work.

whiskeysierra:
> I could have a separate profile or test execution which uses a hard coded spring4 version. That might work.

Mmh. Though about it some more and I don't believe it will work.

A separate test execution means I need to run all tests twice. A separate profile might force a specific spring version, but that would still work after someone changed the dependency version since I'm compatible.

whiskeysierra:
👍

PR #767, opened 2019-09-05
Updated retry docs

### LINKED ISSUE #751: riptide-spring-boot-starter reference incorrect about default retry configuration
While using riptides retry configuration we faced some minor issues with the [reference documentation](https://github.com/zalando/riptide/tree/master/riptide-spring-boot-starter#reference) about some defaults.

## Description
Reference documentation stated that the default `max-duration` (TimeSpan) is not set by default. But [it seems to be a TimeSpan of 5 seconds](https://github.com/zalando/riptide/blob/master/riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/RiptideProperties.java#L81).

Reference documentation stated that the default `max-retries` is not set by default. But [it seems to be 2](https://github.com/jhalterman/failsafe/blob/master/src/main/java/net/jodah/failsafe/RetryPolicy.java#L75).

## Expected Behavior
We expected to have no `max-duration` set to `null/none`.  We expected delays were resolved dynamically by response headers returned by services like `X-RateLimit-Reset`. But most of the time returned values are greater than 5 seconds.

We expected to have `max-retries` not set to behave like infinite number of retries.

## Actual Behavior
`max-duration` (summarized amount of time for all retries) kicks in and makes dynamic resolving delay useless.

`max-retries`: Retries were canceled after second 429 response from service even if service answered again with 429 response.

## Possible Fix
`max-duration`: initialise with `null`

`max-retries`: update reference documentation.

whiskeysierra:
## Description

Docs and implementation were out of sync.

## Motivation and Context

Fixes #751

## Types of changes

## Checklist:

### DIFF
--- a/riptide-spring-boot-autoconfigure/README.md
+++ b/riptide-spring-boot-autoconfigure/README.md
@@ -349,14 +349,14 @@ For a complete overview of available properties, they type and default value ple
 | `│   │   ├── min-size`                  | `int`          | `1`                                              |
 | `│   │   ├── max-size`                  | `int`          | same as `connections.max-total`                  |
 | `│   │   ├── keep-alive`                | `TimeSpan`     | `1 minute`                                       |
-| `│   │   └── queue-size`                | `int`          | `0`                                              |
+| `│   │   └── queue-size`                | `int`          | `0` (unbounded)                                  |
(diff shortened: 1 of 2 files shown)


fatroom:
👍

PR #783, opened 2019-09-12
 Replaced BackupRequestPlugin with BackupRequest policy 

### LINKED ISSUE #770: The bean 'taskScheduler', defined in null, could not be registered.
## Description

I tried running latest riptide version `3.0.0-RC.4`  and springboot version `2.1.8.RELEASE`. However, the Springboot application could not start. It failed with error:

> The bean 'taskScheduler', defined in null, could not be registered. A bean with that name has already been defined in class path resource [org/zalando/riptide/autoconfigure/RiptideAutoConfiguration$SchedulingAutoConfiguration.class] and overriding is disabled.

Possible place of error cause:
https://github.com/zalando/riptide/blob/589544b3863b0b280682875a08c8d5d08e857242/riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/RiptideAutoConfiguration.java#L35

For now I need to enable Bean overriding as a work around. `allow-bean-definition-overriding: true`.

whiskeysierra:
## Motivation and Context

Maintaining a separate module/plugin is not worth it if we already have failsafe that can do most of the heavy lifting. It's especially nice because now we no longer have a need to configure and maintain separate schedulers.
As a side effect also fixes #770 .

## Types of changes

## Checklist:


riptide-failsafe/src/main/java/org/zalando/riptide/failsafe/BackupRequest.java
```diff
@@ -0,0 +1,34 @@
...
+    @Override
+    @SuppressWarnings("unchecked")
+    public PolicyExecutor<Policy<R>> toExecutor(
+            final AbstractExecution execution) {
+        return (PolicyExecutor<Policy<R>>) create(execution);
+    }
+
+    private PolicyExecutor<? extends Policy<R>> create(
+            final AbstractExecution execution) {
+        return new BackupRequestExecutor<R>(this, execution);
```
fatroom: Why this extracted as separate method?

fatroom:
👍

whiskeysierra:
Unchecked casts

PR #797, opened 2019-10-02
Removed meter registry name hack

whiskeysierra:
## Motivation and Context

Hack is no longer needed because we correctly resolve it by type now.

## Types of changes

## Checklist:

### DIFF
--- a/riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/DefaultRiptideConfigurer.java
+++ b/riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/DefaultRiptideConfigurer.java
@@ -4,8 +4,10 @@ import com.google.common.annotations.VisibleForTesting;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.opentracing.Tracer;
 import lombok.AllArgsConstructor;
+import org.springframework.beans.BeanMetadataElement;
 import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.beans.factory.config.BeanDefinition;
+import org.springframework.beans.factory.config.BeanReference;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.beans.factory.config.ConstructorArgumentValues;
 import org.springframework.beans.factory.config.ConstructorArgumentValues.ValueHolder;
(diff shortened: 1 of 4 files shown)


whiskeysierra:
Looks like the issue is that `SimpleMetricsExportAutoConfiguration` has this annotation `@Configuration(proxyBeanMethods = false)`.

> Specify whether `@Bean` methods should get proxied in order to enforce bean lifecycle behavior, e.g. to return shared singleton bean instances even in case of direct `@Bean` method calls in user code. This feature requires method interception, implemented through a runtime-generated CGLIB subclass which comes with limitations such as the configuration class and its methods not being allowed to declare final.
>
>The default is true, allowing for 'inter-bean references' within the configuration class as well as for external calls to this configuration's `@Bean` methods, e.g. from another configuration class. If this is not needed since each of this particular configuration's `@Bean` methods is self-contained and designed as a plain factory method for container use, switch this flag to false in order to avoid CGLIB subclass processing.
>
> Turning off bean method interception effectively processes `@Bean` methods individually like when declared on non-`@Configuration` classes, a.k.a. "`@Bean` Lite Mode" (see `@Bean`'s javadoc). It is therefore behaviorally equivalent to removing the `@Configuration` stereotype.

I don't really understand why they did that, since `SimpleMeterRegistry` holds state and being a proper singleton would make a lot of sense.

whiskeysierra:
Even getting rid of the `SimpleMetricsExportAutoConfiguration` in favor of defining it by hand doesn't help:

```java
@Bean
public SimpleMeterRegistry simpleMeterRegistry() {
    return new SimpleMeterRegistry();
}
```

When I'm debugging it I can see that the method is called multiple times within a single test and hence my plugins and my test get different instances.

When I do this, it works:

```java
private final SimpleMeterRegistry simpleMeterRegistry = new SimpleMeterRegistry();

@Bean
public SimpleMeterRegistry simpleMeterRegistry() {
    return simpleMeterRegistry;
}
```

whiskeysierra:
Might be related to https://github.com/micrometer-metrics/micrometer/issues/513#issuecomment-378755211, since I can see `preInstantiateSingletons` doing this.

whiskeysierra:
I guess `getBeanNamesForType` triggers this.

> Does consider objects created by FactoryBeans, which means that FactoryBeans will get initialized.

whiskeysierra:
I figured it out. `DefaultRiptideConfigurer` uses bean definitions rather than refs, which causes Spring to create inner beans everywhere:

https://github.com/zalando/riptide/blob/90d7a5f1f4432aa4f0e95ff3a0fe7dd95f6c0106/riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/DefaultRiptideConfigurer.java#L73-L96

lukasniemeier-zalando:
:+1:

fatroom:
👍

PR #827, opened 2019-11-18
Workflow test

whiskeysierra:
## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
     - name: Cache
       uses: actions/cache@v1
       with:
-        path: $HOME/.m2
+        path: ~/.m2
         key: m2
     - name: Set up JDK
       uses: actions/setup-java@v1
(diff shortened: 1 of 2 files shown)


fatroom:
As far as I understood you'll need something like this instead of plugin: https://github.com/marketplace/actions/coveralls-github-action

fatroom:
Seems my assumption that Jacoco producing LCOV files was wrong. 
According to https://github.com/trautonen/coveralls-maven-plugin you'll need to provide repoToken as `-DrepoToken=yourcoverallsprojectrepositorytoken` to `mvn coveralls:report` if you're running outside of traffic CI.

whiskeysierra:
Yes, and they don't have proper support for jacoco right now. See https://github.com/coverallsapp/github-action/issues/22.

I don't want to expose the coveralls token publicly.

fatroom:
You can encrypt coverall token, so I'm not sure what your concern about exposing it: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets

whiskeysierra:
I don't have access to the repo settings, otherwise I could set up the coveralls token as a secret and continue to use the coveralls maven plugin.

whiskeysierra:
:+1:

PR #854, opened 2019-12-06
Removed SpanKindSpanDecorator

### LINKED ISSUE #853: Set span.kind tag during creation time already
## Description

The `span.kind` should be set on the builder already, not on the span.

> span.kind: either "client" or "server". It is important to provide this tag **at Span start time**, as it may affect internal ID generation.
>
> https://opentracing.io/specification/conventions/#rpcs

## Expected Behavior

Span kind tag should be set already on the span builder before the span is started.

## Actual Behavior

Span kind is currently set via span decorator on span, after creation. That is too late.

## Possible Fix

- [ ] Remove `SpanKindSpanDecorator` (no real need to customize anyway)
- [ ] Set `span.kind` to `client` statically, unconditionally within `OpenTracingPlugin` directly

## Context

Compliance with OpenTracing semantic conventions.

## Your Environment

* Version used: 3.0.0-RC.9

whiskeysierra:
## Motivation and Context

Fixes #853

## Types of changes

## Checklist:


fatroom:
👍

PR #867, opened 2020-01-09
Separated root pom from riptide-parent

### LINKED ISSUE #866: BOM pulls in Spring BOM version
I am not sure if the following is intended or not:

`riptide-bom` inherits from `riptide-parent`, which pins a whole lot of additional dependencies other than riptide. Among these are the BOM for spring-boot-dependencies and spring-framework.

When using Spring's dependency management plugin and gradle to import riptide's bom it'll override the project's Spring Boot version (and in turn all other dependencies defined by Spring):

```
buildscript {
    dependencies {
        classpath "org.springframework.boot:spring-boot-gradle-plugin:2.2.2.RELEASE"
    }
}

...

dependencyManagement {
    imports {
        mavenBom "org.zalando:riptide-bom:3.0.0-RC.9"
    }
}

...
```

Now `riptide-bom:3.0.0-RC.9` pulls in dependencies declared in `riptide-parent`, which in turn defines spring-boot-dependencies at version 2.2.1.RELEASE. This will then be the version used for the whole project. The same applies for all dependencies defined in the parent pom's `<dependencyManagement>` section.

## Expected Behavior
Riptide-bom defines only riptide dependencies and nothing else.

## Actual Behavior
Riptide-bom defines dependency versions other than riptide. Among these is `spring-boot-dependencies` BOM. This messes the full dependency tree up and is rather difficult to mitigate.

## Possible Fix
1. Do not inherit from `riptide-parent` in `riptide-bom`.
2. Split `riptide-parent` into two poms: one that defines metadata, e.g. `riptide-build`, and another that inherits from it and defines project-wide versions. Then inherit `riptide-bom` from `riptide-build`.
3. Do not add `<dependencyManagement>` section in `riptide-parent`. (Somewhat defeats the purpose of it than :D)

(The same issue applies to logbook-bom. I'm not going to create a ticket there for now.)

Does that make sense? Maybe I am doing something stupid here?!

fatroom:
## Description
This PR introduces org.zalando:riptide pom that serves as root pom of the project.
riptide-bom and riptide-parent are both inherits from this root pom, all other modules depends on riptide-parent. This allows to exclude dependency management influence of the bom, but keeps dependency management in single place for other modules.

## Motivation and Context
Fixes #866 

## Types of changes

## Checklist:


fatroom:
build failed due to PR coming from fork and GH Actions doesn't provide access to secrets in this case. Not sure how to solve this properly

whiskeysierra:
Can you add this?

https://github.com/zalando/problem/pull/189/files

whiskeysierra:
:+1:

PR #882, opened 2020-02-05
Connection leak on failure to parse body

lukasniemeier-zalando:
In case the DefaultMessageReader fails to parse the response body, connections are not returned to the connection pool and stay leased forever.

## Description

A test has been added to show the following scenario:

* the `DefaultMessageReader` fails to parse the response body
* a `RestClientException is raised`
* client connection is not returned to the connection pool and stay leased forever
* connection pool gets exhausted

## Types of changes

## Checklist:

### DIFF
--- a/riptide-core/src/main/java/org/zalando/riptide/DefaultMessageReader.java
+++ b/riptide-core/src/main/java/org/zalando/riptide/DefaultMessageReader.java
@@ -51,6 +51,7 @@ final class DefaultMessageReader implements MessageReader {
                 throw e;
             }
         } catch (final RestClientException e) {
+            // unpack wrapped exception (Spring 5 only)
             propagateIfPossible(e.getCause(), IOException.class, HttpMessageNotReadableException.class);
             throw e;
         }
--- a/riptide-core/src/main/java/org/zalando/riptide/DefaultMessageReader.java
+++ b/riptide-core/src/main/java/org/zalando/riptide/DefaultMessageReader.java
@@ -44,9 +44,13 @@ final class DefaultMessageReader implements MessageReader {
     private <I> I readBody(final Type type, final ClientHttpResponse response) throws IOException {
         final ResponseExtractor<I> extractor = new HttpMessageConverterExtractor<>(type, converters);
         try {
-            return extractor.extractData(response);
+            try {
+                return extractor.extractData(response);
+            } catch (final IOException | RuntimeException e) {
+                response.close();
+                throw e;
+            }
         } catch (final RestClientException e) {
-            response.close();
             propagateIfPossible(e.getCause(), IOException.class, HttpMessageNotReadableException.class);
             throw e;
         }
--- a/riptide-core/src/main/java/org/zalando/riptide/DefaultMessageReader.java
+++ b/riptide-core/src/main/java/org/zalando/riptide/DefaultMessageReader.java
@@ -46,6 +46,7 @@ final class DefaultMessageReader implements MessageReader {
         try {
             return extractor.extractData(response);
         } catch (final RestClientException e) {
+            response.close();
             propagateIfPossible(e.getCause(), IOException.class, HttpMessageNotReadableException.class);
             throw e;
         }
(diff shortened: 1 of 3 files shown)


riptide-core/src/main/java/org/zalando/riptide/DefaultMessageReader.java
```diff
@@ -44,9 +44,13 @@
     private <I> I readBody(final Type type, final ClientHttpResponse response) throws IOException {
         final ResponseExtractor<I> extractor = new HttpMessageConverterExtractor<>(type, converters);
         try {
-            return extractor.extractData(response);
+            try {
+                return extractor.extractData(response);
+            } catch (final IOException | RuntimeException e) {
+                response.close();
```
whiskeysierra: If this fails with its own `IOException` we might loose the original one. Would that be an issue or can we ignore that?

riptide-httpclient/src/test/java/org/zalando/riptide/httpclient/AbstractApacheClientHttpRequestFactoryTest.java
```diff
@@ -179,6 +186,27 @@ void shouldReleaseConnection() {
         });
     }
 
+    @Test
+    void shouldReleaseConnectionOnFailureToReadBody() {
+        driver.addExpectation(onRequestTo("/wrong-content-type"), giveResponse("[]", "text/plain"));
+        driver.addExpectation(onRequestTo("/wrong-content-type"), giveResponse("[]", "text/plain"));
+
+        final Runnable request = throwingRunnable(() -> {
```
whiskeysierra:
```suggestion
        final ThrowingRunnable<Exception> request = () -> {
```

lukasniemeier-zalando: On the one hand "why not" but on the other hand "why"?

riptide-core/src/main/java/org/zalando/riptide/DefaultMessageReader.java
lukasniemeier-zalando: Aha! Another reason `SneakyThrows` must go: I've checked the contract of `ClientHttpResponse` which specifies that implementations must swallow `IOExceptions`, hence I went without another `try-catch` and possibly a "suppressed exception". Enter Riptide's `ApacheClientHttpResponse` which actually throws `IOExceptions` :disappointed: 

IMO we should rather change `ApacheClientHttpResponse` to not throw `IOExceptions`. 

You can still have a case for any `RuntimeException` obviously, but in that case I favor swallowing the original one as this is an unexpected scenario.

riptide-httpclient/src/test/java/org/zalando/riptide/httpclient/AbstractApacheClientHttpRequestFactoryTest.java
whiskeysierra: I usually prefer to use the type over the functions. But it's actually shorter...

riptide-core/src/main/java/org/zalando/riptide/DefaultMessageReader.java
whiskeysierra: > IMO we should rather change `ApacheClientHttpResponse` to not throw `IOExceptions`.

Sounds good

riptide-core/src/main/java/org/zalando/riptide/DefaultMessageReader.java
```diff
@@ -44,7 +44,12 @@
...
+                throw e;
+            }
         } catch (final RestClientException e) {
             propagateIfPossible(e.getCause(), IOException.class, HttpMessageNotReadableException.class);
```
whiskeysierra: Now that we figured out why we have this statement, maybe we should put a comment on it. At least for our future selves.

whiskeysierra:
:+1:

PR #883, opened 2020-02-05
Added mock authorization provider for @RiptideClientTests

whiskeysierra:
## Motivation and Context

As of now, users of `@RiptideClientTest` that had `riptide.clients.example.oauth.enabled: true` would suffer from the following exception during tests:

```
java.nio.file.NoSuchFileException: /meta/credentials/example-token-type
```

Currently there are several workarounds, each with their own drawbacks:

1. `@SpringBootTest(properties="riptide.clients.example.oauth.enabled: false")`
   - needs to be done for each client in each test
2. Register a custom `AuthorizationProvider`
   - needs to implemented (`return () -> "fake";`)
   - needs to be done for each client in each test
3. Change credentials directory and either mount secrets or have static files prepared
   - needs to keep track of file resources
   - needs to be done for each client
 
This pull request still allows to use any of those, but when using `@RiptideClientTest` it will register a mock `AuthorizationProvider` by default that will return a static string. That way, if needed, it can still be overridden, but users will get a less annoying default behavior.

## Types of changes

## Checklist:

### DIFF
--- a/riptide-spring-boot-autoconfigure/README.md
+++ b/riptide-spring-boot-autoconfigure/README.md
@@ -635,7 +635,7 @@ final class RiptideTest {
 }
 ```
 
-**Beware** that all components of a client below and including `ClientHttpRequestFactory` are replaced by mocks.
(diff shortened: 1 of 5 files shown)


fatroom:
👍

lukasniemeier-zalando:
:+1:

PR #887, opened 2020-02-06
Avoid threads being scheduled twice

lukasniemeier-zalando:
## Description

Currently threads are being scheduled twice on their request executor.

* at the very start of a Riptide request (required for isolation)
* after the network phase (intented to force execution back on request
executor)

The second handover was introduced with NIO to accomodate for request
factories which use their own pool. The benefit of forcing execution
back on the request executor is questionable, as any IO is done by
definition after the network phase.

Removing the handover prevents the same excution being scheduled twice
onto the same request executor. In high pressure situations (more
requests than configured threads on a single client), this leads to
deadlock situations in which running executions cannot leave the network
phase, as all other threads are occupied with new executions waiting for
a free slot to start.

## Motivation and Context

We are running in such deadlocks.

## Types of changes

## Checklist:

### DIFF
--- a/riptide-core/src/main/java/org/zalando/riptide/AsyncPlugin.java
+++ b/riptide-core/src/main/java/org/zalando/riptide/AsyncPlugin.java
@@ -20,12 +20,4 @@ final class AsyncPlugin implements Plugin {
                 .thenCompose(identity());
     }
 
-    @Override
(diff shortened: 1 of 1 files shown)


whiskeysierra:
:+1:

PR #888, opened 2020-02-06
Renamed oauth to auth

### LINKED ISSUE #780: Unclear documentation/configuration around AuthorizationProvider in starter
Spring Boot Starter of Riptide 3.0.0-RC.4

The documentation indicates that a bean of type `AuthorizationProvider` can be wired to override the default oauth variant. Unfortunately, this only works when setting `riptide.clients.<ID>.oauth.enabled: true`. The setup of the `AuthorizationPlugin` is conditional behind this property in `DefaultRiptideRegistrar.registerAuthorizationPlugin`.

Setting `oauth.enabled: true` in order to use a non-oauth implementation is misleading and not really straight-forward. I am not sure how it is best solved.

## Possible Implementation
* Rename property to just `auth`; maybe introduce `auth.provider: oauth` or just `auth.oauth: ...` to conditionally setup Zalando's defaults
* Always wire `AuthorizationPlugin` when `AuthorizationProvider` is found; disadvantage is it cannot be disabled in tests
* allow wiring an `AuthorizationPlugin` instead of `AuthorizationProvider` (isn't currently picked up)
* leave as is and require the user to use a the custom plugin feature when using a custom `AuthorizationProvider`

whiskeysierra:
## Motivation and Context

Fixes #780 

## Types of changes

## Checklist:

### DIFF
--- a/riptide-spring-boot-autoconfigure/README.md
+++ b/riptide-spring-boot-autoconfigure/README.md
@@ -166,13 +166,15 @@ Required when `tracing` is enabled.
 </dependency>
 ```
 
-#### OAuth support
+#### Authorization support
 
-Required for `oauth` support.
+Required for `auth` support.
 
 Registers a special AuthorizationProvider that built for Zalando's Platform IAM which provides
 OAuth2 tokens as files in a mounted directory. See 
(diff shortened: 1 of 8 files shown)


lukasniemeier-zalando:
:+1:

fatroom:
👍

PR #891, opened 2020-02-08
Added alternative thread pool semantics

whiskeysierra:
## Motivation and Context

Instead of queuing first, thread pools that use a queue will be scaling first up to maximum pool size before they start queuing.

## Types of changes

## Checklist:


riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/DefaultRiptideRegistrar.java
```diff
@@ -147,19 +143,12 @@ private String registerClientHttpRequestFactory(final String id, final Client cl
     private String registerExecutor(final String id, final Client client) {
         // TODO support maxSize = 0 => direct executor?
```
lukasniemeier: Our idea was to go explicit with `threads.enabled: false` for that. Something for a follow-up?

whiskeysierra:
Yes, I'd separate it

lukasniemeier-zalando:
:+1:

PR #946, opened 2020-06-14
Added CVE-2020-9488 to suppresed list.

fatroom:
This is LOW scored cve in transitive dependency (slf4j-api), can be triggered by specific configuration of it.

## Motivation and Context
Want to suppress it to unlock the builds that crashing due to presence of this CVE.


whiskeysierra:
:+1:

PR #994, opened 2020-09-28
Add explicit RetryPolicy for RetryException

### LINKED ISSUE #993: Double Retries with `transient-fault-detection.enabled = true`
## Description
When using `riptide-spring-boot-autoconfigure` with the following config:

```yaml
riptide:
  clients:
    retry-demo:
      retry:
        enabled: true
        max-retries: 3
      transient-fault-detection:
        enabled: true
      base-url: http://localhost:8088
```

and using `RetryRoute` like so:
```java
  http.get("/test").dispatch(series(), on(SERVER_ERROR).call(RetryRoute.retry())).join();
```

A 5xx response will be retried 6 times (for a total of 7 attempts). Setting `transient-fault-detection.enabled` to `false` resolves the issue.

## Expected Behavior
3 retries are executed

## Actual Behavior
6 retries are executed

## Possible Fix
I _think_ it is related to the configuration of two Policies copying `policy` here:

https://github.com/zalando/riptide/blob/789c2347b8addad143e15038ad64598b8a74fffd/riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/FailsafePluginFactory.java#L119-L125

Both policies contain `handle(RetryException.class)`, thus both policies schedule a retry.

A possible fix could be removing the copying of `policy` and adding a third `RetryRequestPolicy` just for `policy`.

## Steps to Reproduce
I can provide an example project as needed

## Your Environment

* Version used: 3.0.0-RC.14

tobias-bahls:
## Motivation and Context
Fixes #993

## Types of changes

## Checklist:

### DIFF
--- a/riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/FailsafePluginFactory.java
+++ b/riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/FailsafePluginFactory.java
@@ -111,7 +111,6 @@ final class FailsafePluginFactory {
         Optional.ofNullable(config.getJitter())
                 .ifPresent(jitter -> jitter.applyTo(policy::withJitter));
 
-        policy.handle(RetryException.class);
         policy.withDelay(delayFunction());
 
         if (client.getTransientFaultDetection().getEnabled()) {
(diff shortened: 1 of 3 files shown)


fatroom:
👍

fatroom:
Thank you for contribution!

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/FailsafePluginFactory.java
```diff
@@ -122,11 +121,12 @@ public static Plugin createRetryFailsafePlugin(
                     .withPolicy(new RetryRequestPolicy(policy.copy()
                             .handleIf(transientConnectionFaults()))
                             .withPredicate(alwaysTrue()))
+                    .withPolicy(new RetryRequestPolicy(policy.handle(RetryException.class)))
```
whiskeysierra, line 124: Is there a risk here without doing a copy?

tobias-bahls: Do you have something specific in mind? Every other use of `policy` is `copy()`'d, so I think this is fine.

whiskeysierra: Unless the lack of copying is overlooked the next time there is a change needed here.

Issue #953, opened 2020-07-05
JAXB Implementation not found in Java 11 when using Failsafe with Retries & Spring Boot


tobias-bahls:
When using `riptide-soap` together with retries from `riptide-failsafe` in Java 11 on a packaged Spring Boot jar, the call fails with ```javax.xml.bind.JAXBException: Implementation of JAXB-API has not been found on module path or classpath.``` despite the correct JARs being in the classpath.

## Description
In this specific combination:
* `riptide-soap`
* `riptide-failsafe` with retries enabled
* Java 11
* Packaged spring boot jar

The [call](https://github.com/zalando/riptide/blob/f905ba5cc042e78c9df58328f6ef47475166a993/riptide-soap/src/main/java/org/zalando/riptide/soap/SOAPHttpMessageConverter.java#L38) to create a `JAXBContext` with the aforementioned message. If retries are disabled, everything works as expected.

The root cause of the issue seems like that Failsafe is using [`ForkJoinPool.commonPool()`](https://github.com/jhalterman/failsafe/blob/cfbd5179f38e7b2a62b90588c0037f007964194e/src/main/java/net/jodah/failsafe/internal/util/DelegatingScheduler.java#L23-L25) internally, resulting us in hitting https://github.com/spring-projects/spring-boot/issues/15737.

## Possible Fix
I can think of two options:

* Explicitly passing the current class' class loader to `JAXBContext.newInstance()`. This requires us to switch to creating `JAXBContext` with `contextPath` instead of the class to be bound.
* Explicitly supplying an `Executor` to Failsafe that creates threads with the proper context class loader.

I don't know which of these options is better, but I'd be happy to implement the change if there is some consensus.

## Steps to Reproduce
I can provide a sample project if needed

## Your Environment
* Version used: 3.0.0-RC.14

fatroom:
First proposal seems a bit more appropriate from my point of view.

whiskeysierra:
Although the second option is a good idea anyway because the common pool is
a poor choice for io bound workloads.

fatroom:
@whiskeysierra can you give a bit more highlight on why it's "poor choice"?

whiskeysierra:
https://dzone.com/articles/be-aware-of-forkjoinpoolcommonpool

The core issue is, that it's a shared global fixed-size thread pool and IO
usually blocks.

tobias-bahls:
[...]

PR #1086, opened 2021-05-04
Reuse parent classloader for failsafe scheduler threads

buzzlighty:
https://github.com/zalando/riptide/issues/953

Supply failsafe plugin with task decorator to reuse parent class loader for scheduler threads

## Description
PreserveContextClassLoaderTaskDecorator ei

## Motivation and Context
See https://github.com/zalando/riptide/issues/953

Essentially JAXBContext invoked within Failsafe ForkJoinPool will use system class loader, which does not work as expected in combination with Spring Boot (please see related issue).

## Types of changes

## Checklist:

### DIFF
--- a/riptide-failsafe/src/main/java/org/zalando/riptide/failsafe/CompositeTaskDecorator.java
+++ b/riptide-failsafe/src/main/java/org/zalando/riptide/failsafe/CompositeTaskDecorator.java
@@ -1,0 +1,20 @@
+package org.zalando.riptide.failsafe;
+
+import java.util.List;
+import lombok.AllArgsConstructor;
+import net.jodah.failsafe.function.ContextualSupplier;
+
+@AllArgsConstructor
+public class CompositeTaskDecorator implements TaskDecorator {
+
+  private final List<TaskDecorator> decorators;
+
+  @Override
+  public <T> ContextualSupplier<T> decorate(ContextualSupplier<T> supplier) {
+    ContextualSupplier<T> result = supplier;
+    for (TaskDecorator decorator: decorators) {
+      result = decorator.decorate(result);
+    }
+    return result;
+  }
+}
--- a/riptide-failsafe/src/main/java/org/zalando/riptide/soap/PreserveContextClassLoaderTaskDecorator.java
+++ b/riptide-failsafe/src/main/java/org/zalando/riptide/soap/PreserveContextClassLoaderTaskDecorator.java
@@ -1,0 +1,29 @@
+package org.zalando.riptide.soap;
+
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
+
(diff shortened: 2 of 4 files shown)


riptide-failsafe/src/main/java/org/zalando/riptide/soap/PreserveContextClassLoaderTaskDecorator.java
```diff
@@ -0,0 +1,29 @@
...
+import org.apiguardian.api.API;
+import org.zalando.riptide.failsafe.TaskDecorator;
+
+/**
+ * @see <a href="https://github.com/zalando/riptide/issues/953">JAXB + ForkJoinPool</a>
+ */
+@API(status = EXPERIMENTAL)
+public class PreserveContextClassLoaderTaskDecorator implements TaskDecorator {
+
+  private final ClassLoader loader = Thread.currentThread().getContextClassLoader();
```
whiskeysierra, line 15: This might not be the class loader that we're looking for. We create task decorators in the main thread, when we wire up the Riptide client. I think it would be cleaner if the current class loader is read within `decorate` but before the closure that is returned.

riptide-failsafe/src/main/java/org/zalando/riptide/failsafe/CompositeTaskDecorator.java
```diff
@@ -0,0 +1,20 @@
+package org.zalando.riptide.failsafe;
+
+import java.util.List;
+import lombok.AllArgsConstructor;
+import net.jodah.failsafe.function.ContextualSupplier;
+
+@AllArgsConstructor
+public class CompositeTaskDecorator implements TaskDecorator {
```
whiskeysierra, line 8:
```suggestion
private final class CompositeTaskDecorator implements TaskDecorator {
```

whiskeysierra: + a static factory method: `TaskDecorator.composite(TaskDecorator...)` and `TaskDecorator.composite(Iterable<TaskDecorator>)`

riptide-failsafe/src/main/java/org/zalando/riptide/soap/PreserveContextClassLoaderTaskDecorator.java
buzzlighty: That what I thought as well.. I will think about reworking

riptide-failsafe/src/main/java/org/zalando/riptide/failsafe/CompositeTaskDecorator.java
buzzlighty: Will do

whiskeysierra:
Replaced by #1110

Issue #1166, opened 2021-11-03
Opentelemetry support


fatroom:
## Detailed Description
OpenTelemetry is the next iteration of OpenTracing.
While it's possible to report tracing via OpenTelemetry-OpenTracing shim this setup introduced additional layer of abstraction and need to be configured separately by every library user. 

## Possible Implementation
new module riptide-opentelemetry and automatic configuration for spring-starter

fatroom:
Point hinted by @whiskeysierra:

> I'd advice to maintain two separate plugins within the OpenTelemetry module. One for tracing and one for metrics.

PR #1169, opened 2021-11-04
Opentelemetry support 

fatroom:
## Description
Ref #1166 

## Types of changes

## Checklist:


riptide-opentelemetry/src/main/java/org/zalando/riptide/opentelemetry/OpenTelemetryPlugin.java
```diff
@@ -0,0 +1,82 @@
...
+
+        @Nullable
+        final Span span = arguments.getAttribute(internalSpan)
+                                   .orElse(null);
+
+        if (span == null) {
+            return execution.execute(arguments);
+        }
+
+        try (final Scope ignored = span.makeCurrent()) {
```
pc-alves, line 75: This is something we want to do right after creating the span so we'll have the contex available within the thread local scope.

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/OpenTelemetryPluginFactory.java
```diff
@@ -0,0 +1,15 @@
...
+import io.opentelemetry.api.trace.Tracer;
+import org.zalando.riptide.Plugin;
+import org.zalando.riptide.opentelemetry.OpenTelemetryPlugin;
+
+final class OpenTelemetryPluginFactory {
+    private OpenTelemetryPluginFactory() {
+
+    }
+
+    public static Plugin create(final Tracer tracer) {
```
pc-alves, line 12: Just as a heads up, that if we are considering to have the OTel plugin generate traces and metrics, then this method will have to be extended or overloaded to the get `Meter` (the equivalent of the `Tracer`, but for metrics)

whiskeysierra:
How much manual instrumentation is really needed with OpenTelemetry, considering that the agent is instrumenting most if not all libraries that we depend on?

riptide-opentelemetry/src/main/java/org/zalando/riptide/opentelemetry/OpenTelemetryPlugin.java
```diff
@@ -0,0 +1,82 @@
...
+
+    private final Attribute<Span> internalSpan = Attribute.generate();
+
+    public OpenTelemetryPlugin(final Tracer tracer) {
+        this.tracer = tracer;
+        this.propagator = GlobalOpenTelemetry.getPropagators().getTextMapPropagator();
+    }
+
+    @Override
+    public RequestExecution aroundAsync(@Nonnull final RequestExecution execution) {
```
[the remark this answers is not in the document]
whiskeysierra: You're correct. The method itself is called from the calling thread. The returned lambda is called on the IO thread though.

The OpenTracing plugin should give you a good blueprint for how to transfer the span over.

riptide-opentelemetry/src/main/java/org/zalando/riptide/opentelemetry/OpenTelemetryPlugin.java
whiskeysierra: Not necessarily. Activating a span only really works in non-reactive systems. Unless you have an alternative implementation of the ScopeManager or whatever it's called in OTel now.

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/OpenTelemetryPluginFactory.java
whiskeysierra: I'd advice to maintain two separate plugins within the OpenTelemetry module. One for tracing and one for metrics.

You should also consider not to support the metrics part of OTel at all, since micro meter is already an established standard that can serve as a bridge towards OTel.

riptide-opentelemetry/src/main/java/org/zalando/riptide/opentelemetry/OpenTelemetryPlugin.java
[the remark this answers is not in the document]
fatroom: You're correct. 
a) Method called from initial thread, so it's proper place to grab parent context. 
Still creation of spans in the method isn't right place, since backup / retry requests wouldn't be covered. 
b) not sure but I want to look on tagging retries/backup requests etc

fatroom:
@whiskeysierra the idea to provide basic semantic instrumentation + option for users to set op names and additional attributes if they want. Just in case agent is not in use.

whiskeysierra:
That would be weird since OpenTelemetry almost relies on the agent
instrumentation for almost 100% of libraries and frameworks

pc-alves:
> How much manual instrumentation is really needed with OpenTelemetry, considering that the agent is instrumenting most if not all libraries that we depend on?

@whiskeysierra Similar to the effort with Riptide's OpenTracing module and the opentracing-toolbox project, we are looking to add Zalando's standard attributes and logs. The other reason to not use the agent is that the instrumentations provided by OTel would fit with Level 1 of the [Maturity Levels](https://github.com/zalando/opentracing-toolbox/blob/main/docs/maturity-levels.md#%EF%B8%8F-level-1-take-it-or-leave-it), "Take it or leave it". There is little possibility to augment, redact, or customize the tracing data.

whiskeysierra:
> > How much manual instrumentation is really needed with OpenTelemetry, considering that the agent is instrumenting most if not all libraries that we depend on?
> 
> @whiskeysierra Similar to the effort with Riptide's OpenTracing module and the opentracing-toolbox project, we are looking to add Zalando's standard attributes and logs. The other reason to not use the agent is that the instrumentations provided by OTel would fit with Level 1 of the [Maturity Levels](https://github.com/zalando/opentracing-toolbox/blob/main/docs/maturity-levels.md#%EF%B8%8F-level-1-take-it-or-leave-it), "Take it or leave it". There is little possibility to augment, redact, or customize the tracing data.

I'd be careful with treating OpenTelemetry similar to OpenTracing.

OpenTracing was/is:
- a specification
- a minimal set of semantic conventions (too limited, one could argue)
- not governing any widespread instrumentation code (instrumenation was left to library/framework and application owners)

OpenTelemetry on the other hand is:
- a bigger, broader specification
- a broad set of semantic conventions (can't judge whether they are enough)
- governing a very sophisticated instrumentation code repository (the majority being agent-exclusive, the rest is done as a reusable library)

OpenTelemetry implements their own, broader set of semantic conventions consistently and it tries to move most/all of the instrumenation code into the agent's monorepo.

Both take a very different approach.

whiskeysierra:
> Again, the agent instrumentation would be perfectly fine if we could get some 'hooks' where we could make some changes, or add some attributes to make traces more in line with what we're looking for.

So, this then? https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/manual-instrumentation.md#creating-spans-around-methods-with-withspan

pc-alves:
The `WithSpan` is definitely something useful, but using it to overcome the limitations of the automatic instrumentation means that developers need to write more instrumentation code, and likely having to restructure the code to meet the instrumentation requirements. For example:

```
void someMethod() {

  /* more code */

  Response response1 = httpclient.request("/some/url", POST, headers, body);

  Response response2 = httpclient.request("/another/url", GET, headers);

  /* doing something with the responses above */
}
```

Assuming an instrumented HTTP client, the code above would create the spans for the HTTP requests with whatever default semantics are included. If the code owners would like to customize the spans for either or both of the HTTP requests they can't. To use the `WithSpan` annotation to counter that, they'd have to extract the call into a separate method, when all that was needed was to add a new attribute.
This also leads to a higher verbosity in tracing data, consuming more resources in services.

The closest thing that the OTel instrumentation provides to what I was describing are [extensions](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/examples/extension/README.md#i-want-to-edit-some-attributes-that-dont-depend-on-any-db-connection-instance). But even those are somewhat limited.
Riptide's instrumentation has a very nice balance between providing as much out-of-the-box telemetry, while allowing users to customize that telemetry. Either through the SpanDecorators, or through call site attributes. It is this balance and ease of use that we are looking for when providing instrumentation for applications.

riptide-opentelemetry/README.md
```diff
@@ -0,0 +1,117 @@
...
+| `exception.type`             | `ErrorSpanDecorator`               | `SocketTimeoutException`          |
+| `exception.message`          | `ErrorSpanDecorator`               | `Socket timed out`                |
+| `exception.stacktrace`       | `ErrorSpanDecorator`               | `SocketTimeoutException at [...]` |
+| `http.host`                  | `HttpHostSpanDecorator`            | `www.github.com`                  |
+| `http.method`                | `HttpMethodSpanDecorator`          | `POST`                            |
+| `http.path`                  | `HttpPathSpanDecorator`            | `/users/{user_id}`                |
+| `http.status`                | `HttpStatusCodeSpanDecorator`      | `200`                             |
+| `peer.hostname`              | `PeerHostSpanDecorator`            | `www.github.com`                  |
+| `retry`                      | `RetrySpanDecorator`¹              | `true`                            |
+| `retry_number`               | `RetrySpanDecorator`¹              | `3`                               |
```
pc-alves, line 69:
```suggestion
| `retry`                      | `RetrySpanDecorator`²              | `true`                            |
| `retry_number`               | `RetrySpanDecorator`²              | `3`                               |
```

riptide-opentelemetry/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+- Span context injection into HTTP headers of requests
+- Extensible span decorators for attributes
+- Seamless integration with [Riptide: Failsafe](../riptide-failsafe)
+
+## Dependencies
+
+- Java 8
+- Riptide Core
+- [OpenTelemetry Java API](https://opentelemetry.io/docs/instrumentation/java/)
```
pc-alves, line 28: Being completely accurate, shouldn't we also mention `opentelemetry-semconv`?

riptide-opentelemetry/README.md
```diff
@@ -0,0 +1,117 @@
...
+| `retry_number`               | `RetrySpanDecorator`¹              | `3`                               |
+|                              | `StaticTagSpanDecorator`¹          | `zone=aws:eu-central-1a`          |
+
+¹ The `CompositeSpanDecorator` allows to treat multiple decorators as one. 
+² **Not** registered by default.
+
+### Span Decorators
+
+Span decorators are a simple, yet powerful tool to manipulate the span, i.e. they allow you to add attributes to the spans. 
+The default set of decorators can be extended by using `OpenTracingPlugin#withAdditionalSpanDecorators(..)`:
```
pc-alves, line 78: This method doesn't actually exist.

```suggestion
The default set of decorators can be extended by using `OpenTelemetryPlugin#withAdditionalSpanDecorators(..)`:
```

riptide-opentelemetry/src/main/java/org/zalando/riptide/opentelemetry/OpenTelemetryPlugin.java
```diff
@@ -0,0 +1,107 @@
...
+public class OpenTelemetryPlugin implements Plugin {
+
+    public static final Attribute<String> OPERATION_NAME = Attribute.generate();
+
+    private final Tracer tracer;
+    private final TextMapPropagator propagator;
+    private final SpanDecorator spanDecorator;
+
+    public OpenTelemetryPlugin(final Tracer tracer, SpanDecorator... decorators) {
+        this.tracer = tracer;
```
pc-alves, line 42: I'm struggling a bit with this. In OpenTracing the Tracer was the center of everything (span creation, span context propagation, span exporting, ...). That's no longer the case with OpenTelemetry. Which is why you then use `GlobalOpenTelemetry` to get the Propagators, for example. In fact, Tracers are supposed to tag the source of their instrumentation. That would mean something like `this.tracer = GlobalOpenTelemetry.getTracerProvider.get("riptide-opentelemetry");`

That being the case I see a couple of options:
1. `OpenTelemetryPlugin` has an empty constructor, and another with SpanDecorators, and we get everything from the GlobalOpenTelemetry.
2. We support a builder to allow users to pass their own Tracer and propagator.

WDYT?

riptide-opentelemetry/src/main/java/org/zalando/riptide/opentelemetry/OpenTelemetryPlugin.java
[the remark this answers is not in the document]
pc-alves: Ok, so if I understand this correctly (and I acknowledge I may be missing something), we are not passing the span to the thread executing the request, rather opting to work with the span and the request's response that is returned from that thread. Did I get it right?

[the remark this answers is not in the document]
fatroom: Almost. We're not passing span from caller method, but context. Spans will be created in the threads that will execute calls

riptide-opentelemetry/src/main/java/org/zalando/riptide/opentelemetry/OpenTelemetryPlugin.java
fatroom: I went from spring perspective which doesn't recommend static initialisation, but injection of dependencies. Still looking after some time after I wrote this I'm not sure why I went with tracer by injection, but static init of propagators. 
Maybe it worth to go with OpenTelemetry instance as constructor parameter and take tracer and propagators out of this instance? Is this approach fits to telemetry principles?

riptide-opentelemetry/README.md
fatroom: I think it would be overkill. Think about this like "we depending on spring boot", but not specifying that we depend on spring-context, spring-web-starter etc

pc-alves: Fair enough.

riptide-opentelemetry/src/main/java/org/zalando/riptide/opentelemetry/OpenTelemetryPlugin.java
pc-alves: I get the injection argument. Certainly helps with testing.
But passing an `OpenTelemetry` will not really change much, from a user perspective. It's still not the user setting the propagators or the tracer. On the other hand, it should be riptide calling `getTracer(instrumentationName, instrumentationVersion)` in order to set the two parameters.

Let's go ahead with the `OpenTelemetry` as a constructor parameter

riptide-opentelemetry/src/main/java/org/zalando/riptide/opentelemetry/OpenTelemetryPlugin.java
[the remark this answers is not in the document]
pc-alves: Looking at the `trace` method the context is used to create the span wrapping the HTTP call (line 68). That span then becomes the current span (line 76), and that context is injected into the headers (line 77), which are then added to `arguments` (line 80).
This means that the executing thread has access to the context through those headers, and if needed will have to build the context with them. I guess I was expecting something like we have in the [OpenTracing plugin](https://github.com/zalando/riptide/blob/main/riptide-opentracing/src/main/java/org/zalando/riptide/opentracing/OpenTracingPlugin.java#L174), that passes the span itself (although, with OpenTelemetry it would make sense to just pass the context. WDYT?

riptide-opentelemetry/src/main/java/org/zalando/riptide/opentelemetry/OpenTelemetryPlugin.java
```diff
@@ -0,0 +1,107 @@
...
+import org.zalando.riptide.opentelemetry.span.SpanDecorator;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.BiConsumer;
+import javax.annotation.Nonnull;
+
+public class OpenTelemetryPlugin implements Plugin {
```
pc-alves, line 33: Right now we can only extend the set of span decorators, but we cannot override the default set like we can with the OpenTracing plugin's [`withSpanDecorators`](https://github.com/zalando/riptide/blob/main/riptide-opentracing/src/main/java/org/zalando/riptide/opentracing/OpenTracingPlugin.java#L150).

riptide-opentelemetry/src/main/java/org/zalando/riptide/opentelemetry/span/ErrorSpanDecorator.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.http.client.ClientHttpResponse;
+import org.zalando.riptide.RequestArguments;
+
+import java.io.IOException;
+import java.util.function.Predicate;
+
+import static org.zalando.fauxpas.FauxPas.throwingPredicate;
+
+public class ErrorSpanDecorator implements SpanDecorator {
+    private final Predicate<ClientHttpResponse> predicate;
```
pc-alves, line 14: The way this is setup I don't think you really need the predicate. It is set by the constructor, with no way for the user to override it. In the OpenTracing plugin users can override the predicate for the [`ErrorSpanDecorator`](https://github.com/zalando/riptide/blob/main/riptide-opentracing/src/main/java/org/zalando/riptide/opentracing/span/ErrorSpanDecorator.java) (it is annotated with `@AllArgsConstructor`). Either we go with something along the same lines, or we make this one simpler by removing the predicate. I would favor consistency across Plugins, and follow the same approach.

riptide-opentelemetry/src/main/java/org/zalando/riptide/opentelemetry/span/StaticSpanDecorator.java
```diff
@@ -0,0 +1,21 @@
...
+import org.zalando.riptide.RequestArguments;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class StaticSpanDecorator implements SpanDecorator {
+    private final Map<AttributeKey<String>, String> attributes = new HashMap<>();
+
+    public StaticSpanDecorator(Map<String, String> source) {
+        source.forEach((k, v) -> attributes.put(AttributeKey.stringKey(k), v));
```
pc-alves, line 14: This could result in an NPE. I'm fine, in principle, as this would fail during application startup. Just double checking that this was a conscious decision.

fatroom: my assumption was that as the method isn't annotated as expecting nullable parameter here, it's okay to skip null checks and throw NPE

riptide-opentelemetry/src/main/java/org/zalando/riptide/opentelemetry/OpenTelemetryPlugin.java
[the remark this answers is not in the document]
fatroom: OpenTracing plugin just adds layer of abstraction by providing two kinds of injectors, headers and noop. But effectively injection to headers happens during execution of trace method via `injection.inject(tracer, arguments, context)`. Current implementation doesn't have support for such injectors, and I'm not sure if needed.

riptide-opentelemetry/src/main/java/org/zalando/riptide/opentelemetry/OpenTelemetryPlugin.java
fatroom: Added

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/OpenTelemetryPluginFactory.java
pc-alves: When this OTel [PR with a Micrometer bridge](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4919) gets merged we can just document how to use it in order for the client to emit metrics in OTel format.

riptide-opentelemetry/src/main/java/org/zalando/riptide/opentelemetry/span/StaticSpanDecorator.java
pc-alves: Again, fine by me. My comment was just a safety check.

riptide-opentelemetry/src/test/java/org/zalando/riptide/opentelemetry/OpenTelemetryPluginTest.java
```diff
@@ -0,0 +1,252 @@
...
+
+        final List<SpanData> spans = otelTesting.getSpans();
+
+        assertThat(spans, hasSize(2));
+
+        assertThat(spans.get(1).getSpanId(), is(parent.getSpanContext().getSpanId()));
+
+        final SpanData child = spans.get(0);
+        assertThat(child.getParentSpanId(), is(parent.getSpanContext().getSpanId()));
+        assertThat(child.getStatus(), is(StatusData.error()));
```
pc-alves, line 161: Just for completeness, I'd recommend verifying that we're also adding details about the exception from the call to `recordException(...)` in the `ErrorSpanDecorator`
We can look for one of the attributes or events listed [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/exceptions.md).

riptide-spring-boot-autoconfigure/README.md
```diff
@@ -236,6 +236,17 @@ Required when `propagate-flow-id` is enabled.
 </dependency>
 ```
 
+#### `Telemetry`
```
pc-alves, line 239:
```suggestion
#### Telemetry
```

pc-alves: Should we rename the section in line 210 from `Tracing` to `OpenTracing` ? Just to separate one from the other, as they are both related to Tracing.

riptide-spring-boot-autoconfigure/README.md
```diff
@@ -300,9 +311,15 @@ riptide:
         tags:
           peer.service: example
         propagate-flow-id: true
+      telemetry:
+        enabled: true
+        tags:
```
pc-alves, line 316: To be consistent with the standard, this should be named `attributes` instead of `tags`

pc-alves: `attributes` is what was named in the `RiptideProperties.Telemetry` class. Guess this was just a typo.

```suggestion
        attributes:
```

riptide-spring-boot-autoconfigure/src/test/java/org/zalando/riptide/autoconfigure/OpenTelemetryTestAutoConfiguration.java
```diff
@@ -0,0 +1,14 @@
+package org.zalando.riptide.autoconfigure;
+
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.api.trace.TracerProvider;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class OpenTelemetryTestAutoConfiguration {
```
pc-alves, line 9: Do we still need this class after the change to the OpenTelemetryPlugin constructor?

riptide-spring-boot-autoconfigure/src/test/resources/application-default.yml
```diff
@@ -63,6 +63,10 @@ riptide:
         keystore:
           path: example.keystore
           password: password
+      telemetry:
+        enabled: true
+        tags:
```
pc-alves, line 68: This should be `attributes` as well.

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/ValueConstants.java
```diff
@@ -14,4 +14,6 @@ private ValueConstants() {}
     static final String METER_REGISTRY_REF = "\n\t\t\n\t\t\n\uE000\uE001\uE001\n\t\t\t\t\n";
     static final String TRACER_REF = "\n\t\t\n\t\t\n\uE000\uE001\uE002\n\t\t\t\t\n";
     static final String LOGBOOK_REF = "\n\t\t\n\t\t\n\uE000\uE001\uE003\n\t\t\t\t\n";
+    static final String TELEMETRY_TRACER_REF = "\n\t\t\n\t\t\n\uE000\uE001\uE004\n\t\t\t\t\n";
```
pc-alves, line 17: We don't need this anymore, now we're getting the tracer from the `OpenTelemetry` instance.

riptide-opentelemetry/src/test/java/org/zalando/riptide/opentelemetry/OpenTelemetryPluginTest.java
fatroom: Added event verification

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/ValueConstants.java
fatroom: missed during clean up step

fatroom:
👍

fatroom:
👍

pc-alves:
👍

PR #1252, opened 2022-03-29
Reuse spring classloader for failsafe scheduler threads

buzzlighty:
Contributes to https://github.com/zalando/riptide/issues/953

## Description
Fix `PreserveContextClassLoaderTaskDecorator` in order to set a class loader from invoking thread to a runner thread.

## Motivation and Context
Previously this task decorator has been created to resolve above mentioned issue. However there's a bug in this task decorator which prevents it from working as expected. The code in labmda should assign class loader to runner thread but it was assigning it to invoking thread, which does not make sense as it's the same thread which the class loader is being obtained from.

## Types of changes

## Checklist:


fatroom:
@buzzlighty can you please rebase on master to get the build passing?

buzzlighty:
@fatroom sure, can you have another look please?

fatroom:
@buzzlighty will do today

fatroom:
👍

fatroom:
Thank you for your contribution, included in release 3.2.1

PR #1296, opened 2022-07-07
OpenTelemetry: the http.status_code wasn't set if the route was incomplete

zeitlinger:
OpenTelemetry: the http.status_code wasn't set if the route was incomplete


fatroom:
👍

PR #1297, opened 2022-07-07
OpenTelemetry: provided dependencies

zeitlinger:
OpenTelemetry: set the dependencies to "provided" to avoid problems with incompatible API changes in opentelemetry


zeitlinger:
breaks when running in javaagent - dropping the PR for now

PR #1384, opened 2023-01-19
fix incompatibility with java agent

zeitlinger:
when running in java agent, the downstream http client is also instrumented
therefore it's important that the Scope is activated while we call the downstream client


whiskeysierra:
Activating a scope effectively sets a thread local value.
For that to work you got to stick to a single thread, which Riptide isn't.
Your test uses mocks and doesn't really create a realistic usage.

PR #1387, opened 2023-01-20
add OpenTelemetry metrics

zeitlinger:
Try out how to add OpenTelemetry metrics.

It seems a bit redundant to reimplement all metrics using otel API, so trying to leverage the existing bridge.

I'm ussure how to best integrate the MicrometerPlugin. I would create a composite Plugin, but this doesn't work, because users call the OpenTelemetryPlugin constructor directly.

### DIFF
--- a/riptide-opentelemetry/pom.xml
+++ b/riptide-opentelemetry/pom.xml
@@ -18,13 +18,17 @@
 
     <properties>
         <opentelemetry.version>1.22.0</opentelemetry.version>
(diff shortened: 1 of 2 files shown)


whiskeysierra:
I'd separate the tracing and metrics aspects into two separate
modules/plugins. OpenTelemetry will support logs in the future as well.
Which would just be another plugin/module.

If there is a desire to have all combined, you can create a dedicated
module for that, which just bundles the others under the hood.

fatroom:
I believe we discussed the topic of different modules in the past when telemetry plugin was introduced.

@zeitlinger check #1169

Issue #1392, opened 2023-01-31
Consolidate usage of http mock servers in tests


fatroom:
## Context
Currently the different tests in the code base use MockRestServiceServer from spring framework as well as RestDriver (same named library). 
RestDriver is based on jetty 9 which is incompatible with the spring framework 6. 
For moving forward we need to avoid direct relation on java.servlet based packages.

## Possible Implementation
Keeping #837 in mind it's better to avoid usage of MockRestServiceServer in the tests, so we can use use MockWebServer from [OkHttp](https://github.com/square/okhttp/tree/master/mockwebserver).
The library is independent from javax.server namespace problems, as well as providing good enough level control. 

Another option would be to wait for wiremock 3 release (beta version of wire mock 3 already available, though not sure if that would be possible without simultaneous migration of the spring framework to version 6 due to JEE api changes)

Semernitskaya:
## Mock servers comparison approach
I've compared suggested implementations. 
   
I've collected main use cases of `RestDriver` in the project tests (e.g. `verify()`, respond with delay, use custom `org.apache.http.client.HttpClient`), wrote a test with main use cases using `RestDriver` - see [HttpMockTestClientDriver](https://github.com/zalando/riptide/pull/1437/files#diff-9d3eb8d0977d210dcee3b726f6148ba0eeedd34ede18662b9679472cfda3fa7c), then I tried to rewrite this test using different mock servers implementations.
   
**Notes:**
1. Test is located in `riptide-failsafe` module as I'm better familiar with it from my previous task and many use cases are based on tests from `riptide-failsafe` (e.g. processing delayed response)
2. After rewriting, every test case can be executed successfully, and the same logic is verified.

## Mock servers comparison results
### Spring MockRestServiceServer
### OkHttp MockWebServer
### WireMock WireMock class

Semernitskaya:
@fatroom I agree, yes sure - I'll continue working on this task. My plan:
- Migrate modules one by one
- Send draft PR when first couple of modules are migrated to make sure that I'm on the right track 
- Delete MockSetup classes (they are used for Spring MockRestServiceServer initialisation)
- Keep Spring MockRestServiceServer in spring related modules (riptide-spring-boot-starter, riptide-spring-boot-autoconfigure)

PR #1441, opened 2023-04-11
Use mock http servers from okhttp

Semernitskaya:
## Description

- All tests in the project that used `com.github.restdriver.clientdriver.ClientDriver` as mock web server were migrated to `okhttp3.mockwebserver.MockWebServer` instead.
- `MockWebServerUtil` util class was introduced to simplify migration and add functionality missing in the new `okhttp3.mockwebserver.MockWebServer`.
- `MockWebServerUtil` class is duplicated across modules and can be reviewed only in one place. The same duplication approach was used for the existing `MockSetup` class. In the future we can avoid copying and pasting them by introducing a new project module that will be used only with the test scope (see discussion https://github.com/zalando/riptide/pull/1441#discussion_r1162980272).
- `com.github.restdriver.clientdriver.ClientDriver` allows you to define response for a **specific** request and then use `verify()` method to ensure that driver got this exact request. For `okhttp3.mockwebserver.MockWebServer` you usually define common response for all possible requests, then check the request you actually got (using `server.takeRequest(...)` method). Multiple `verify()` methods were added to `MockWebServerUtil` class to simplify such check. See https://github.com/zalando/riptide/pull/1441/files#diff-c2b11cfb53a03add49f5f20b1efdbd3c238230a2ec547eed41218d6c8dbdfff6R95 for a simple example of old validation vs new validation. 

## Motivation and Context

Fixes https://github.com/zalando/riptide/issues/1392.    
Migration to the new mock test server is required because `com.github.restdriver.clientdriver.ClientDriver` doesn't support latest Jetty version and thus blocks migration to the Spring Boot 3 & Spring 6.
Different alternatives were analysed and discussed in the task, see https://github.com/zalando/riptide/issues/1392#issuecomment-1484743482 for more details.

## Types of changes

## Checklist:

- [ ] My change requires a change to the documentation. (Not applicable as changes are in test classes only)
- [ ] I have updated the documentation accordingly. (Not applicable as changes are in test classes only)
- [ ] I have added tests to cover my changes. (Not applicable as changes are in test classes only)


Semernitskaya:
I've migrated 5 modules in this draft PR:
riptide-failsafe, riptide-faults (required only removal of `rest-client-driver` dependency), riptide-httpclient, riptide-idempotency (required only removal of `rest-client-driver` dependency), riptide-logbook

riptide-failsafe/src/test/java/org/zalando/riptide/failsafe/MockWebServerUtil.java
```diff
@@ -0,0 +1,50 @@
+package org.zalando.riptide.failsafe;
```
Semernitskaya, line 1: I've collected common util methods in `MockWebServerUtil` class, currently this class is copied to the every module where I used `okhttp3.mockwebserver.MockWebServer` (the same approach is used for the existing class `MockSetup`). We can avoid copying and pasting it by introducing new project module that will be used only with test scope but I'm not sure if it is worths investing time into

fatroom: I would recommend to skip extracting separate module, totally fine with copying paste in tests

riptide-compression/src/test/java/org/zalando/riptide/compression/RequestCompressionPluginTest.java
```diff
@@ -17,94 +19,97 @@
...
 
     @ParameterizedTest
     @ArgumentsSource(RequestFactorySource.class)
     void shouldCompressRequestBody(final ClientHttpRequestFactory factory) {
-        driver.addExpectation(onRequestTo("/")
-                        .withMethod(POST)
-                        .withHeader("X-Content-Encoding", "gzip") // written by Jetty's GzipHandler
-                        .withBody(equalTo("{}"), "application/json"),
-                giveResponse("", "text/plain"));
+        server.enqueue(textMockResponse(""));
```
Semernitskaya, line 56: We cannot use Jetty's `GzipHandler` anymore, so we cannot test header inserted by it. I've tested headers inserted by the plugin and that body was actually compressed (by decompressing it and checking result)

fatroom:
Thank you for your contribution!

fatroom:
👍

Issue #1351, opened 2022-12-09
Is there a plan to upgrade to Spring boot 3 & Spring 6?


KeshavRathi:
Is there a plan to upgrade to Spring boot 3 & Spring 6

## Context
A project of our uses Riptide and we wanted to upgrade the Spring version in it but can't as Riptide still uses the previous version and has few deprecated classes.

fatroom:
We evaluating upgrade to spring boot 3, but no established timeline for a moment

fatroom:
Steps needs to be performed to complete migration:

- [ ] switch to jdk17
- [ ] switch to apache components >= 5.1
- [ ] switch to Jakarta EE >= 9
- [ ] if compatible with Jakarta ee rest-client version exists switch to it, otherwise switch to okhttp mock server for testing
- [ ] accommodate spring changes in sync http client version

Open points:
1. deprecate async Http (?) / find the way to work with reactive spring? maybe outside of the basic spring 6 migration
2. evaluate interface based http clients and decide if can be supported?

PR #1471, opened 2023-06-17
Migration to Spring Boot 3

msdousti:
Migration to Spring Boot 3 along with `apache components >= 5.1` and `Jakarta EE >= 9`.

## Motivation and Context
Resolves https://github.com/zalando/riptide/issues/1351

## Types of changes

## Checklist:
    - Aligned with @cberg-zalando to update the documentation.


riptide-httpclient/src/test/java/org/zalando/riptide/httpclient/BufferingApacheClientHttpRequestTest.java
```diff
@@ -0,0 +1,23 @@
...
+    @Test
+    void shouldThrowIllegalArgumentException() throws URISyntaxException {
+        final HttpUriRequest httpUriRequest = mock(HttpUriRequest.class);
+        when(httpUriRequest.getUri()).thenThrow(URISyntaxException.class);
+
+        final BufferingApacheClientHttpRequest request = new BufferingApacheClientHttpRequest(null, httpUriRequest);
+        assertThrows(IllegalArgumentException.class, request::getURI);
+    }
+
+}
```
fatroom, line 23: side note - please add new line at the end of the file

riptide-idempotency/src/test/java/org/zalando/riptide/idempotency/MethodOverrideIdempotencyDetectorTest.java
```diff
@@ -0,0 +1,33 @@
...
+                Map.of("X-HTTP-Method-Override", Collections.singletonList("BAD_METHOD"))
+        );
+
+        final MethodOverrideIdempotencyDetector methodOverrideIdempotencyDetector
+                = new MethodOverrideIdempotencyDetector();
+
+        assertThat(methodOverrideIdempotencyDetector.test(arguments, test), is(Decision.NEUTRAL));
+    }
+
+}
```
fatroom, line 33: new line at the end of the file missing here

riptide-parent/pom.xml
```diff
@@ -17,19 +17,26 @@
...
         <fauxpas.version>0.9.0</fauxpas.version>
-        <jackson.version>2.14.2</jackson.version>
-        <jackson-databind.version>2.14.2</jackson-databind.version>
-        <junit-jupiter.version>5.9.3</junit-jupiter.version>
-        <logbook.version>3.0.0</logbook.version>
-        <micrometer.version>1.10.4</micrometer.version>
-        <mockito.version>4.11.0</mockito.version>
+        <jackson.version>2.15.2</jackson.version>
+        <jackson-databind.version>2.15.2</jackson-databind.version>
+        <junit-jupiter.version>5.10.0-M1</junit-jupiter.version>
```
fatroom, line 22: any specific reason to use milestone release? is there something that you've used that isn't available in junit 5.9?

cberg-zalando:
Any reason why you went with the commit 7052bff at all given that the project uses dependabot?

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/OpenTracingFlowIdAutoConfiguration.java
```diff
@@ -0,0 +1,55 @@
...
+    @Bean
+    public Flow flow(final Tracer tracer) {
+        return Flow.create(tracer);
+    }
+
+    @API(status = INTERNAL)
+    @Configuration
+    @ConditionalOnClass(HttpClient.class)
+    @ConditionalOnMissingBean(FlowHttpRequestInterceptor.class)
+    @ConditionalOnProperty(name = "opentracing.flowid.httpclient.enabled", havingValue = "true", matchIfMissing = true)
```
fatroom, line 44: matchIfMissing option should be false here.

But generally I'm concerned that http client tries now to assume that it's working in http servlet environment and establishing classes on incoming layer of the app.

cve-suppressions.xml
```diff
@@ -1,67 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
-    <suppress>
```
fatroom, line 3: Thanks for fixing this list! Was in my plans to do it someday!

cberg-zalando:
> @cberg-zalando I'm fine with using default formatting of IDEA here. So star imports and re-ordering are acceptable.

I just wanted to have a judgement call.

riptide-httpclient/src/test/java/org/zalando/riptide/httpclient/BufferingApacheClientHttpRequestTest.java
msdousti: Don't know why my IDE didn't add new line - I've configured it to do so:
image

Anyway, used this trick to fix it:
https://unix.stackexchange.com/a/161853/500398

```bash
git ls-files -z | while IFS= read -rd '' f; do if file --mime-encoding "$f" | grep -qv binary; then tail -c1 < "$f" | read -r _ || echo >> "$f"; fi; done
```

riptide-idempotency/src/test/java/org/zalando/riptide/idempotency/MethodOverrideIdempotencyDetectorTest.java
msdousti: Fixed for all files, as per previous comment.

riptide-httpclient/src/test/java/org/zalando/riptide/httpclient/BufferingApacheClientHttpRequestTest.java
whiskeysierra: May I suggest to use [.editorconfig](https://editorconfig.org/) for which there are IDE plugins available. Good chance contributors already have it installed.

Same goes for [.pre-commit-config.yaml](https://pre-commit.com/hooks.html) which among other useful hooks (newline at the end of each file, etc.) also supports editorconfig.

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/OpenTracingFlowIdAutoConfiguration.java
msdousti: > matchIfMissing option should be false here.

The tests won't pass if it's set to false - and I guess it will break existing Riptide users because they may not have set `opentracing.flowid.httpclient.enabled` by default.

We can instead enable it with either `riptide.defaults.tracing.enabled = true` or `riptide.clients.*.tracing.enabled = true` being true (the latter having a wildcard makes it a bit hard 🤔).

> But generally I'm concerned that http client tries now to assume that it's working in http servlet environment and establishing classes on incoming layer of the app.

To my understanding, it doesn't make this assumption. However, if it works in such environment, Spring gets the hint how to order auto-configurations.

The current version of Riptide works in the same way - it just imports this class from `opentracing-flowid-starter`:
https://github.com/zalando/opentracing-toolbox/blob/main/opentracing-flowid/opentracing-flowid-autoconfigure/src/main/java/org/zalando/opentracing/flowid/autoconfigure/OpenTracingFlowIdAutoConfiguration.java

Since that project is in maintenance mode, I thought the only way to keep OpenTracing is add this little configuration.

cve-suppressions.xml
msdousti: My pleasure! I really like that with this upgrade, it is now very minimalistic :)

fatroom:
@cberg-zalando can you update MIGRATION.md with changes from this PR?

cberg-zalando:
> @cberg-zalando can you update MIGRATION.md with changes from this PR?

Will do when I have the time available. When I started the migration, I had more time available. But I guess, you know how things can go.

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/OpenTracingFlowIdAutoConfiguration.java
cberg-zalando: I think, it was always a bit weird that the FlowId topic was linked to Open Tracing. But I think, this was the issue when the original tracer project was started.

While it has something to do with tracing in general, in the end it is `just` an interceptor fiddling with the `X-Flow-Id` header. So given that we want to release a major version, we should think about whether it makes sense to decouple it and just use the `tracing.propagate-flow-id` property for it?

msdousti: I agree that it can be an improvement to decouple it, but it falls out the scope of this PR, which is "just" migration to SB 3.😊 

We can have another task in the milestone of Release 4.0 to apply the suggested decoupling. The rationale behind not doing so is to give this PR a chance to being released as soon as possible.

With Logbook, we did multiple release candidates, used them in our services, found the issues, and fixed them until we could make the new major release. I can imaginge we'll do the same, and in the meantime we have the chance to work on a decopling.

fatroom:
This looks fine for me now. 

fatroom:
:+1:

PR #1484, opened 2023-07-03
Fix cache auto configuration

msdousti:
Using Riptide without caching enabled must not require `httpclient5-cache`.

## Description
* Fixed `RiptideProperties` such that it does not require `org.apache.hc.client5.http.impl.cache.CacheConfig`.
* Fixed `org.springframework.boot.autoconfigure.AutoConfiguration.imports` to have separate configs for main and test (otherwise, production code would require `MockRestServiceServer` to be in class path)
* Updated README.md to use `httpclient5-cache` instead of `httpclient-cache`.

## Types of changes

## Checklist:

### DIFF
--- a/riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/RiptideProperties.java
+++ b/riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/RiptideProperties.java
@@ -4,10 +4,11 @@ import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.Setter;
-import org.apache.hc.client5.http.impl.cache.CacheConfig;
+import org.apache.hc.core5.util.TimeValue;
 import org.apiguardian.api.API;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.NestedConfigurationProperty;
+import org.springframework.stereotype.Component;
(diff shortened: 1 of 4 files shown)


fatroom:
👍

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/RiptideProperties.java
```diff
@@ -35,6 +36,7 @@
 @Setter
 @NoArgsConstructor
 @AllArgsConstructor
+@Component
```
cberg-zalando, line 39: Interesting that you need to add this annotation as `@ConfigurationProperties` should normally make this available to Spring.

msdousti: Intellij showed an error saying `not registered via @enableconfigurationproperties or marked as spring component.`

The project compiled anyway, but I followed the advice here:
https://stackoverflow.com/a/57950415/459391

> This is expected as @ConfigurationProperties does not make a class a Spring Component. Mark the class with @Component and it should work. Note that a class can only be injected if it is a Component.

cberg-zalando: Thanks for the insights. Yes, I remember that you have to add this annotation to your main class for example, to get this working. I guess, given that this is a library, it makes sense to follow that approach.

PR #1528, opened 2023-11-02
Fixing ConnectException name

yuppie-flu:
## Description
This PR fixes the docs by using the correct name for the JRE [ConnectException](https://docs.oracle.com/javase/8/docs/api/java/net/ConnectException.html) class. See also [this usage](https://github.com/zalando/riptide/blob/5f004a2e3ad920c87ae393a13157c31363c1ebd3/riptide-faults/src/main/java/org/zalando/riptide/faults/TransientFaults.java#L76) in the source code.

### DIFF
--- a/riptide-faults/README.md
+++ b/riptide-faults/README.md
@@ -56,7 +56,7 @@ new RetryRequestPolicy(
 ```
 
 The default considers the following exceptions to be transient connection faults:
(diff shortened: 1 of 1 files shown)


fatroom:

fatroom:
👍

Issue #1671, opened 2024-11-18
Riptide 4 does not respect X-RateLimit-Reset header when used through riptide-spring-boot-autoconfigure


sklv-max:
## Description
Upon receiving a 429 response along with X-RateLimit headers a client should obey those headers. 
For instance, if `X-RateLimit-Remaining` header has 0 as its value, then a client should wait `X-RateLimit-Reset` seconds before making another retry. 

However, when riptide clients are configured through `riptide-spring-boot-autoconfigure` in riptide 4.1.0, those headers get ignored. 
This was not the case in Riptide 3.x.x

## Expected Behavior
X-RateLimit headers are respected

## Actual Behavior
X-RateLimit headers are ignored

## Possible Fix
I am not sure but one possible reason is this [change](https://github.com/zalando/riptide/commit/6bf111c9dec8c8d23e501e19d0bac2838917d142#diff-689625fc30ba207613b49e4baf732c8d247a0a265b67baa8fb5a29096245eee6R41).

Riptide autoconfiguration [creates](https://github.com/zalando/riptide/blob/e4caf8352402a938a29dc494e889a506e8d4d661/riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/FailsafePluginFactory.java#L186) a composite delay function, consisting of two functions: `RetryAfterDelayFunction` and `RateLimitResetDelayFunction`. When `Retry-After` is missing and the `X-RateLimit ` headers are present it should take result of the `RateLimitResetDelayFunction` (see [here](https://github.com/zalando/riptide/blob/e4caf8352402a938a29dc494e889a506e8d4d661/riptide-failsafe/src/main/java/org/zalando/riptide/failsafe/CompositeDelayFunction.java#L31)). But since `RetryAfterDelayFunction` does not return null anymore, its result is taken (`Duration.ofMinutes(-1)`). Because it is a negative interval, retry happens immediately. 

However, this is just a guess.

## Steps to Reproduce

The following test works in Riptide 3.4.0 (with JDK 8) but does not work in the current main branch (JDK 17)
```
package org.zalando.riptide.autoconfigure.retry;

import static org.hamcrest.MatcherAssert.assertThat;
import static org.hamcrest.Matchers.greaterThanOrEqualTo;
import static org.springframework.http.HttpStatus.Series.SUCCESSFUL;
import static org.springframework.test.web.client.ExpectedCount.times;
[...]

PR #1672, opened 2024-11-20
Fix X-RateLimit headers handling when using CompositeDelayFunction

sklv-max:
1. Filtering out default durations in the CompositeDelayFunction.
2. Returning default duration from the CompositeDelayFunction instead of null.

## Description
Since RateLimitResetDelayFunction and RetryAfterDelayFunction are returning `Duration.ofMinutes(-1)` instead of null as the default value since Riptide 4, it is needed to check not only for non-null values but also if there are any non-default delays in the CompositeDelayFunction.

## Motivation and Context
1. In Riptide 4 the default value in RateLimitResetDelayFunction and RetryAfterDelayFunction was changed from null to `Duration.ofMinutes(-1)` [link](https://github.com/zalando/riptide/commit/6bf111c9dec8c8d23e501e19d0bac2838917d142)
2. Failsafe 2.x.x used both negative duration and null as default values ([link](https://javadoc.io/static/net.jodah/failsafe/2.4.4/index.html?net/jodah/failsafe/function/DelayFunction.html)) 
3. Failsafe 3.x.x mentions only negative values as default values [(link)](https://failsafe.dev/javadoc/core/dev/failsafe/DelayablePolicyBuilder.html#withDelayFn-dev.failsafe.function.ContextualSupplier-). If null is returned from delay function, it leads to NPE in the Failsafe.

As Riptide 4, if CompositeDelayFunction consists of two functions (A and B) and A returns default value and B returns a positive meaningful value, the default value would be taken since it is not null.
 
The issue is opened [here](https://github.com/zalando/riptide/issues/1671).

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X] I have added tests to cover my changes.

### DIFF
--- a/riptide-failsafe/src/main/java/org/zalando/riptide/failsafe/CompositeDelayFunction.java
+++ b/riptide-failsafe/src/main/java/org/zalando/riptide/failsafe/CompositeDelayFunction.java
@@ -29,8 +29,9 @@ public final class CompositeDelayFunction<R> implements ContextualSupplier<R, Du
                     }
                 })
                 .filter(Objects::nonNull)
+                .filter(delay -> !Duration.ofMinutes(-1).equals(delay))
                 .findFirst()
-                .orElse(null);
+                .orElse(Duration.ofMinutes(-1));
     }
 
     @SafeVarargs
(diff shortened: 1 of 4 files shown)


fatroom:
@sklv-max please verify your gpg key and sign commit: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

sklv-max:
> @sklv-max please verify your gpg key and sign commit: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

Thank you for pointing out, have done that.

fatroom:
👍

puncleV:
👍

fatroom:

PR #1682, opened 2025-01-05
Update Spring to 6.2

yuppie-flu:
## Description

This PR updates Spring Framework version to 6.2.1 and replaces removed (deprecated) APIs to alternatives:
* removed `AbstractClientHttpResponse` abstract class was replaced to `ClientHttpResponse` interface

In Spring Framework 6.2 `org.springframework.http.HttpRequest` interface received a new method: `Map<String, Object> getAttributes()` without a default implementation. The PR doesn't implement the method and throw UnsupportedOperationException instead. It seems that this method is needed only for very rare use cases (like passing attributes to interceptors, see https://github.com/spring-projects/spring-framework/issues/32027). However, we can discuss if some implementation in Riptide makes sense.

## Motivation and Context

In Spring Framework 6.2 release several [deprecated APIs](https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-6.2-Release-Notes#removed-apis) were removed. The current Riptide code uses some of them, in particular: `org.springframework.http.client.AbstractClientHttpResponse`.
The newest Spring Boot 3.4 release uses Spring Framework 6.2. The incompatibility of the current Riptide with Spring Framework 6.2 makes it impossible to upgrade services that use Riptide to the latest Spring Boot. ClassNotFoundException about `AbstractClientHttpResponse` is thrown at runtime.

## Types of changes

## Checklist:
- [?] My change requires a change to the documentation.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -77,7 +77,7 @@
         <maven.compiler.source>17</maven.compiler.source>
         <maven.compiler.target>17</maven.compiler.target>
 
-        <spring.version>6.1.14</spring.version>
+        <spring.version>6.2.1</spring.version>
     </properties>
 
     <build>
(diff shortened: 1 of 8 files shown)


fatroom:
👍

yuppie-flu:
:+1:

PR #1700, opened 2025-02-19
Bumped opentelemetry-semconv to the latest version

cberg-zalando:
## Description
Latest Obversaviblity SDK 3.1.13 contains a bump of `opentelemetry-semconv`from 1.28.0-alpha to 1.30.0-rc.1 introducing a breaking change as the `io.opentelemetry.semconv.SemanticAttributes` class was removed and replaced with more specific sub class in the sub package `incubating`.

Nevertheless, looking at the class it had the deprecation marker since version 1.24.0 (Riptide uses 1.25-alpha currently), so I guess, it is the usual issue that these deprecation warnings are overlooked until it is too late.

## Motivation and Context
It allows projects using Obversaviblity SDK to use the latest version.

## Types of changes

### DIFF
--- a/riptide-opentelemetry/pom.xml
+++ b/riptide-opentelemetry/pom.xml
@@ -18,7 +18,7 @@
 
     <properties>
         <opentelemetry.version>1.39.0</opentelemetry.version>
-        <opentelemetry-semconv.version>1.25.0-alpha</opentelemetry-semconv.version>
+        <opentelemetry-semconv.version>1.30.0-rc.1</opentelemetry-semconv.version>
     </properties>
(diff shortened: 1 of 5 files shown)


fatroom:
The non existing classes are still used (i.e. HttpStatusCodeSpanDecorator)

cberg-zalando:
> The non existing classes are still used (i.e. HttpStatusCodeSpanDecorator)

Yeah, somehow I overlooked this one class.

fatroom:
👍

zDimi:
:+1:

PR #1702, opened 2025-02-21
Remove Maven OWASP dependency check plugin

cberg-zalando:
As seen in the latest build of #1700, the downloading of the CVE database in the OWAP dependency-check plugin takes a lot of time in case you don't have a specific key. As also another Zalando repository was affected by this, the recommendation is to remove it and rely on on GitHub's dependabot integration.


yuppie-flu:
It would be nice to remove the `NVD_API_KEY` variable [here](https://github.com/zalando/riptide/blob/610c3f17d85508daeab6fc1d01035be5f314934c/.github/workflows/build.yaml#L19) as well.

fatroom:
Thank you for contribution. Included in #1703

Issue #1721, opened 2025-03-21
Accept threadpool executors for FailsafePlugin through yaml configuration.


rameshmalla:
The proposal is to support yaml based configuration to add threadpool exuectors to failsapfe plugin

## Detailed Description
riptide spring boot starter by default, accepts most HTTP client configurations through YAML, making it simple and convenient to configure and quickly start using the Riptide client. For the FailsafePlugin, the current approach for introducing a custom thread pool executor involves configuring it through the codebase.

While this approach offers greater control over thread pool management and allows for fine-tuning of execution threads, it requires additional code-based configuration as opposed to the more declarative YAML-based configuration used for other HTTP client settings. 

The proposal is to introduce YAML based thread pool configuration for FailSafePlugin and re-use the existing thread pool construction logic in ripitde which supports opentracing and micrometer features out of the box.

PR #1722, opened 2025-03-21
Accept threadpool executors for FailsafePlugin through yaml configuration

rameshmalla:
See https://github.com/zalando/riptide/issues/1721

## Types of changes

## Checklist:


riptide-spring-boot-autoconfigure/README.md
```diff
@@ -27,11 +27,23 @@ riptide.clients:
       enabled: true
       fixed-delay: 50 milliseconds
       max-retries: 5
+      threads:
+        max-size: 10
+        min-size: 2
+        enabled: true
```
lukasniemeier-zalando, line 33: If I set this to `false` (as a user), I'd expect failsafe to use the calling thread here (i.e. no custom thread pool at all) - but actually it would use the (default) fork-join-pool then, right? This might be a bit confusing for users.

riptide-spring-boot-autoconfigure/README.md
```diff
@@ -27,11 +27,23 @@ riptide.clients:
...
+        keep-alive: 5 minutes
+        queue-size: 10
```
lukasniemeier-zalando, line 35: do these two necessarily make sense for the use cases of failsafe?

riptide-spring-boot-autoconfigure/README.md
rameshmalla: I believe this was earlier discussed and therefore it was added to the README https://github.com/zalando/riptide/tree/main/riptide-failsafe#custom-executor

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/DefaultRiptideRegistrar.java
```diff
@@ -138,27 +140,29 @@ private String registerBaseURL(final String id, final Client client) {
         });
     }
 
-    private Object createExecutor(final String id, final Client client) {
-        if (client.getThreads().getEnabled()) {
-            return ref(registerExecutor(id, client));
+    private Object createExecutor(final String id, final String metricName, final Client client, @Nullable final RiptideProperties.Threads threads) {
+        if (threads != null && threads.getEnabled()) {
```
fatroom, line 144: How threads can become null?

fatroom:
@rameshmalla please make sure that your commits has verified signature. check the GH documentation on setting it up: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

fatroom:
@rameshmalla I see you enabled signature, you need to sign commit 162c4c86d26b58fe61510a5d1defc1be3de09613 as well.

rameshmalla:
> @rameshmalla I see you enabled signature, you need to sign commit [162c4c8](https://github.com/zalando/riptide/commit/162c4c86d26b58fe61510a5d1defc1be3de09613) as well.

I think I finally did it 😄

riptide-micrometer/src/main/java/org/zalando/riptide/micrometer/ThreadPoolMetrics.java
```diff
@@ -14,11 +13,9 @@
 import static com.google.common.collect.ImmutableList.copyOf;
 import static io.micrometer.core.instrument.binder.BaseUnits.TASKS;
 import static io.micrometer.core.instrument.binder.BaseUnits.THREADS;
-import static lombok.AccessLevel.PRIVATE;
 import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 
 @API(status = EXPERIMENTAL)
-@AllArgsConstructor(access = PRIVATE)
```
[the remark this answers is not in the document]
rameshmalla: My bad, was debugging without lombok, reverted it back.

riptide-spring-boot-autoconfigure/README.md
rameshmalla: These settings aren't strictly required for Failsafe, but they are standard properties of a typical thread pool. I can remove them from the YAML configuration. However, since I'm reusing the existing "Threads" config, users can still set these properties if needed—and I believe at that point, they know what they're doing.

Or do you have other suggestions @fatroom @lukasniemeier-zalando  ?

riptide-spring-boot-autoconfigure/src/main/java/org/zalando/riptide/autoconfigure/DefaultRiptideRegistrar.java
rameshmalla: By default, Failsafe policies use `null` for threads (unlike client HTTP threads), so I’ve added a null check to handle that case safely.

fatroom:
👍

PR #1754, opened 2025-06-25
Migrate to Central Portal

lukasniemeier-zalando:
## Description

Migrating from OSSRH to Central Portal for publishing of releases. See https://central.sonatype.org/publish/generate-portal-token/ for authentication. 

## Motivation and Context
See https://central.sonatype.org/faq/what-is-different-between-central-portal-and-legacy-ossrh/ 

- [x] fix `pom` violations
- [x] fix javadocs violations

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -228,17 +228,6 @@
         </pluginManagement>
     </build>
 
-    <distributionManagement>
-        <snapshotRepository>
-            <id>ossrh</id>
-            <url>https://oss.sonatype.org/content/repositories/snapshots</url>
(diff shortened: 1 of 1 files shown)


lukasniemeier-zalando:
I have done a successful test deployment by running `mvn deploy` locally. All validations were passed ✅ I have dropped the deployment before it was published. In https://github.com/zalando/riptide/pull/1754/commits/2e6e4a2d2aff4aeb4deeeb793aa35c2431dc4de3 we have now activated auto-publishing for future (true) releases.

PR #1845, opened 2026-04-04
Clarify retry semantics, fault types, and transient-fault-detection requirement

### LINKED ISSUE #952: Document retry mechanic in a more detailed way
The retry mechanic of riptide is only briefly described in the documentation and could have more details to cover different questions.

## Detailed Description

Especially the note that only idempotent requests are retried should be documented as this cost me quite some time to figure out. Additionally you could document that by default only connection errors are retried and otherwise you should define the retry behavior yourself. The riptide-failsaife documentation reads like the configuration of the retry behavior in the config would be good enough.

## Context
It will prevent people from making mistakes that could be prevented with a couple of sentences in the documentation.

sri-adarsh-kumar:
Clarifies retry semantics around idempotency, fault types, and response-based retries in riptide-failsafe and spring-boot-autoconfigure READs.
Fixes socket-fault retry docs to correctly require transient-fault-detection instead of being standalone.
Adds consolidated resilience docs overview at docs/resilience.md.

Closes https://github.com/zalando/riptide/issues/952

### DIFF
--- a/docs/resilience.md
+++ b/docs/resilience.md
@@ -67,16 +67,17 @@ riptide.clients:
       jitter: 25 milliseconds
 ```
 
-**What `retry:` config controls:** timing, backoff, jitter, and limits. It does **not** make
-Riptide retry arbitrary `4xx` or `5xx` responses automatically.
+**What `retry:` config controls:** timing, backoff, jitter, and limits. It does **not** by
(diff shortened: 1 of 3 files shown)


fatroom:
👍

cberg-zalando:
:+1:

PR #1846, opened 2026-04-04
Add riptide-example-basic onboarding module

### LINKED ISSUE #1031: The README is not enough documentation to actually start using this library
## Description

There seems to be some information missing in the README to actually use the library.

## Expected Behavior

Reading the README (or some other documentation linked from there) should be enough to actually start using the library, i.e. sending out HTTP requests and doing something with the response.

## Actual Behavior

I've read the README (and several of the other documents linked from there), and still have no clue on what exactly to do.
Reading the source of some of the classes helped somewhat, but not fully.

## Possible Fix

* Make sure there is example code which is complete (i.e. including all needed imports).
* If any other dependencies are needed (i.e. some of the ~ 20 riptide-* projects), make this visible.
* List/link the relevant classes/methods I need to use, or might want to use.
* If there is some other documentation which already has this↑ (maybe a Javadoc site?), make sure it's clearly visibly linked.

## Context

I remembered from some presentations at work (at Zalando) that "Riptide" is the new way of doing HTTP requests (and getting responses) nowadays. I got a (private) problem where I want to automate filling a form to book an appointment, for which I need some HTTP requests (and handling the responses), so I said "let's try this out".

## Steps to Reproduce

Being on vacation, I don't have any access to production code which is using Riptide (from which I could copy this), and I also didn't find any example code in this repository.
I'm just looking at the README (and everything which is linked from there).

The first step in my project was

> Call this URL (with `GET`) and when it returns a redirect (any 3xx, though in this case a 302), do something with the URL in the Location header.

I've set up an `Http` object using the riptide-spring-boot-starter.

```java
@Autowired
@Qualifier("formularserver")
private Http formularServer;
```

Looking at the [Requests](https://github.com/zalando/riptide#requests) section, sending the request seems easy:

```java
formularServer
    .get(INITIAL_URL)
```

Now [handling the response](https://github.com/zalando/riptide#responses) ...
This short section with one conceptual paragraph, and one piece of example code seems to be all there is on documentation for what is supposed the *main feature* of Riptide. And the example code doesn't even work when copying it, because all the imports are missing (specifically, the static imports for `on`, `series()`, `contentType()`, `status()`, `anyStatus()` and the Spring HTTP status code and series enum values).
(It also references user code, but I can understand this.)

With some browsing through the classes in riptide-core, I found that I need those imports (I guess `*` imports would also work):

```java
import static org.springframework.http.HttpStatus.Series.REDIRECTION;
import static org.zalando.riptide.Bindings.on;
import static org.zalando.riptide.Navigators.series;

...

public void findForm() {
	formularServer
		.get(INITIAL_URL)
		.dispatch(series(),
				on(REDIRECTION)
				    .dispatch( /*TODO: now what? */ null, null));
}
```
... or possibly one of the `.call(...)` in the last line?

After some more digging I found that I can just implement my own Route, or rather a `ThrowingConsumer<ClientHttpResponse>` which will just fetch the header from the response itself.

```java
public void findForm() {
	formularServer
		.get(INITIAL_URL)
		.dispatch(series(),
				on(REDIRECTION).call(this::printLocation)
				);
}
void printLocation(ClientHttpResponse response) {
	LOG.info("location header: ", response.getHeaders().get("Location"));
}
```
I still have the suspicion that there is some higher-level way of doing this, though, which I didn't find yet.

I didn't yet get to the next step of actually using this location header for making the next request, and then extracting data from the response to that request.

(Actually running this (with a debugger) shows that the Apache HTTP client which is used here already takes care of the redirect internally, so I guess this all was just wasted effort.)

## Your Environment

* Version used: library version 3.0.0-RC.15, README version of ec6b614.

sri-adarsh-kumar:
Adds riptide-example-basic: a runnable onboarding module demonstrating routing, redirect header extraction, and body deserialization via the Spring Boot starter path.
Enhances README.md with a Getting Started section; excludes the module from Maven Central. Fixes CI build failure where duplicate-finder-maven-plugin failed on reactor builds due to missing sibling target/classes at validate phase.
Closes #1031

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,16 +25,41 @@ protocol and Java. Riptide allows users to leverage the power of HTTP with its u
  - **Upgrading from 3.x to 4.x?** Please refer to the [Migration Guide](MIGRATION.md#riptide-40-migration-guide).
  - **Upgrading from 2.x to 3.x?** Please refer to the [Migration Guide](MIGRATION.md#riptide-30-migration-guide).
 
+## Getting Started
+
+New to Riptide? Start here:
+
+1. **Pick your dependency** – see the table below.
+2. **Run the [basic example](riptide-example-basic)** – a self-contained module with complete imports and tests.
+3. **Read [docs/concepts.md](docs/concepts.md)** – explains routing trees, navigators, and bindings.
+
+### Dependency quick-reference
+
+| Use case | Dependency |
+|---|---|
+| Minimal / manual client | `riptide-core` |
+| Spring Boot application | `riptide-spring-boot-starter` |
+| Retries, circuit breaker, timeouts | `riptide-failsafe` |
+| Transient fault detection (socket errors) | `riptide-faults` |
+
 ## Example
 
 Usage typically looks like this:
 
 ```java
+import static org.springframework.http.HttpStatus.Series.SUCCESSFUL;
+import static org.zalando.riptide.Bindings.on;
+import static org.zalando.riptide.Navigators.series;
+import static org.zalando.riptide.Types.listOf;
+
 http.get("/repos/{org}/{repo}/contributors", "zalando", "riptide")
     .dispatch(series(),
-        on(SUCCESSFUL).call(listOf(User.class), users -> 
+        on(SUCCESSFUL).call(listOf(User.class), users ->
             users.forEach(System.out::println)));
 ```
+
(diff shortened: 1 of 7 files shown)


fatroom:
👍

cberg-zalando:
:+1:

